PR #16, opened 2016-08-24
Moving OTS Test to Pajarito

ccoffrin:
removing the use of Gurobi and using Pajarito instead.


ccoffrin:
@mlubin, how about 90% coverage.  Thanks to Pajarito!

mlubin:
:+1:

PR #17, opened 2016-08-26
Cleaning up ots tests to not require NLP models

ccoffrin:
Tests will fail until the next version of Pajarito.


ccoffrin:
Retest when this is merged, https://github.com/mlubin/Pajarito.jl/pull/110

PR #22, opened 2016-09-01
Linting src/core directory

kersulis:
Cosmetic changes in src/core directory. Mostly removing trailing spaces and blank lines. Functionality should be unaltered: the only syntax changes are e2ae40b (rename `gen` to `g` to prevent confusion with outer scope `gen`) and d02c15a (rewrite to avoid declaring unused var `l`).

### DIFF
--- a/src/core/base.jl
+++ b/src/core/base.jl
@@ -91,7 +91,7 @@ function JuMP.solve(pm::GenericPowerModel)
     try
         solve_time = getsolvetime(pm.model)
     catch
-        warn("there was an issue with getsolvetime() on the solver, falling back on @timed.  This is not a rigorous timing value.");
+        warn("there was an issue with getsolvetime() on the solver, falling back to @timed.  This is not a rigorous timing value.")
     end
 
     return status, solve_time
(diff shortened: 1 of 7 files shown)


src/core/base.jl
```diff
@@ -154,18 +143,16 @@ function build_sets(data :: Dict{AbstractString,Any})
     gen_idxs = collect(keys(gen_lookup))
     branch_idxs = collect(keys(branch_lookup))
 
-
-    buspair_indexes = collect(Set([(i,j) for (l,i,j) in arcs_from]))
-    buspairs = buspair_parameters(buspair_indexes, branch_lookup, bus_lookup)  
+    buspair_indexes = collect(Set([a[2:3] for a in arcs_from]))
+    buspairs = buspair_parameters(buspair_indexes, branch_lookup, bus_lookup)
```
ccoffrin: Can you tell me your motivation to move from pattern matching to array indexing in this case?

kersulis: Sure, Lint.jl pointed out that `l` was defined but not used, so I tried to make it unnecessary. `[(i, j) for (NaN, i, j) in x] also works. @mlubin, is there any performance advantage to getting rid of `l`? I do think it looked better before my change.

mlubin: No, there's no real reason to remove ``l`` there. I'd say it's a false positive from the linter.

kersulis: Good to know, I'll revert the change when I'm at my laptop again.

kersulis:
Tried to squash 12 commits to 4, ended up with 17. Redoing this PR with a new branch, thanks for your patience.

PR #23, opened 2016-09-02
src/core cleanup

kersulis:
Take 2. This time I didn't change variable names or replace pattern matching with indexing. I did re-word a couple comments and fixed a typo in a warning message, like before.


ccoffrin:
Looks great.  Thanks!

PR #25, opened 2016-09-02
Suppress warnings during testing

kersulis:
[Logging.jl](https://github.com/kmsquire/Logging.jl) can override `info()` and `warn()` during testing to suppress all but the test summary messages. This makes the output of `Pkg.test()` easier to read.

After this change, any use of `info()` in test files will be ignored, so `println` must be used instead. I didn't see any `info()` calls in the test files, so this is just something to keep in mind when writing future test files.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -6,3 +6,4 @@ GLPKMathProgInterface
 Pajarito
 
 BaseTestNext
+Logging
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,4 +1,7 @@
 using PowerModels
+using Logging
+# suppress warnings during testing
+Logging.configure(level=ERROR)
 
 using Ipopt
 using Pajarito


ccoffrin:
This looks great.  You are correct info() is not yet used in the code.

OT, but do you know what the `WARNING: could not attach metadata for @simd loop.` messages are about?

kersulis:
I'm not seeing those messages, but that might be because I'm on v0.4.6 still.

This does raise an important point: if new warning messages crop up in the future, this PR will make them invisible during testing. Why don't I change warnings like these:

```julia
WARNING: there was an issue with getsolvetime() on the solver, falling back on @timed.  This is not a rigorous timing value.
WARNING: this code only supports angmin values in -89 deg. to 89 deg., tightening the value on branch 1 from -360.0 to -60 deg.
WARNING: this code only supports positive rate_a values, changing the value on branch 1 from 0.0 to 1804.6048863043677
```
to use `info()` instead, then change `Logging.configure(level=ERROR)` to `Logging.configure(level=WARNING)`?

ccoffrin:
Locally some times I get them, not sure why.  They do show up on the travis build for this PR if you are curious.

I am fine with both info and warnings being silent during package tests.  If the existence of the info/warning is very important, then special tests should be added to verify that function.  Although this feature is not currently supported in Julia, as far as I know.

To your point of using warning vs info.  I am open to the change, but I think it depends a little on the conventions in the Julia community about what types of messages belong in the two streams.

I originally set these to warning because I wanted users think hard, before running experiments that have these warnings.

PR #28, opened 2016-09-28
minor change to sad opf model for improved performance

ccoffrin:


mlubin:
"sad" seems to be an unfortunate name for the model

ccoffrin:
Ha!  Too true.  This problem is really an artifact of making interesting test cases, and will eventually become obsolete. 

I plan to move this model it out of this code into the test generation code base.

PR #31, opened 2016-09-29
#30, fix Julia 0.5 warnings

### LINKED ISSUE #30: Fix Julia v0.5 Warnings

kersulis:
Fixes #30:
* replace [...=>...for...] with Dict(...=>...for...)
* replace readall with readstring
* replace symbol with Symbol

After changes, no warnings are printed during pre-compilation, and Pkg.test() output is clean on 0.5.

All other changes are due to my text editor automatically removing trailing whitespace and whitespace on empty lines.

### DIFF
--- a/src/io/matpower.jl
+++ b/src/io/matpower.jl
@@ -24,7 +24,7 @@ function parse_matrix(lines, index)
     matrix_body_lines = [matrix_assignment_rhs]
     found_close_bracket = contains(matrix_assignment_rhs, "]")
 
-    while index + line_count < last_index && !found_close_bracket 
+    while index + line_count < last_index && !found_close_bracket
         line = strip(lines[index+line_count])
 
         if length(line) == 0 || line[1] == '%'
@@ -73,7 +73,7 @@ end
 
 
 function parse_matpower(file_string)
-    data_string = readall(open(file_string))
+    data_string = readstring(open(file_string))
     data = parse_matpower_data(data_string)
 
     for branch in data["branch"]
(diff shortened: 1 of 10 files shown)


ccoffrin:
It looks like the new Dict syntax is not backward compatible with v0.4.

What options do we have to support both?

kersulis:
Whoops, sorry for not catching that.

Looks like avoiding warnings on both 0.4 and 0.5 involves [rewriting the statements as for loops](https://github.com/JuliaLang/Compat.jl/issues/231) unfortunately. Which if I'm not mistaken would convert this:

```julia
bp_angmin = Dict(bp => -Inf for bp in buspair_indexes)
bp_angmax = Dict(bp =>  Inf for bp in buspair_indexes)
bp_line = Dict(bp => Inf for bp in buspair_indexes)
```
to this:
```julia
bp_angmin = Dict()
bp_angmax = Dict()
bp_line = Dict()
for bp in buspair_indexes
    bp_angmin[bp] = -Inf
    bp_angmax[bp] = Inf
    bp_line[bp] = Inf
end
```
I don't like that nearly as much, but glancing through other packages it appears that's what people are doing.

kersulis:
Another way that works across both:
```julia
bp_angmin = Dict([(bp, -Inf) for bp in buspair_indexes])
```
So just replacing the `a=>b` syntax with `(a,b)`.

kersulis:
Okay, I rewrote the `Dict()` statements so each argument is a vector of 2-tuples.

Then I noticed `readstring()`, the 0.5 replacement for `readall()`, isn't defined in 0.4, so 0.4 tests still weren't passing. So I added [Compat.jl](https://github.com/JuliaLang/Compat.jl) to the requirements and put `@compat` in front of the 2 lines that call `readstring()`. All this does is replace `readstring` with `readall` for 0.4 users.

src/io/matpower.jl
```diff
@@ -73,7 +73,7 @@ end
 
 
 function parse_matpower(file_string)
-    data_string = readall(open(file_string))
+    @compat data_string = readstring(open(file_string))
```
mlubin: You don't need ``@compat`` here. ``readstring`` is just a method exported by the Compat module

kersulis: Good to know, I'll revert those 2 changes.

ccoffrin:
Looks good to me.  Thanks a bunch!

src/io/matpower.jl
kersulis: Just curious: I used a force push to get rid of those changes, so now when I click "View changes" above your comment, the previous commit is gone. Is there a better way to modify a commit in an PR? I hesitate to add another commit just to undo a bit of the last one...

PR #33, opened 2016-10-13
Implementation of expansion planning

rb004f:
@ccoffrin I have implemented expansion planning as an algorithm for PowerModels.jl A few points for discussion

1. If you are so inclined, please review the expansion planning SOC model.  It is passing all my tests, but it is the place where I could see making an error, given the need to mixed on/off constraints on lines

2. I have everything "custom" living in expansion.jl. If you think we should move that out to the model packages (wr.jl, etc.), we can do that.

3. The expansion modeling almost identically implements the ots model. I didn't use the ots directly as 1) this would preclude having and ots + expansion model and 2) expansion planning does not assume all lines are buildable (i.e. not all lines can be switched. To avoid duplicate code, what do you think about breaking the ots constraints to have helper functions where the on/off variable is passed as an argument? That way 90% of the code is reused, except for which on/off variable gets passed to it (switching or expansion).

4. constraint_complex_voltage bothers me... it is a line based constraint, so I would love it to have an argument where the branch is passed to it.  I understand why you didn't do this... in the wr model, you have wr/wi variables defined by bus pair in opf, but have to do it on a line basis in ots... I had to do some ugly things to that function in expansion planning see

```
constraint_complex_voltage_expansion_on_off{T <: AbstractWRForm}(pm::GenericPowerModel{T})
```
maybe you have some thoughts on how to do this

5. I went down the route to calling different constraint functions depending on if the expansion variable was present or not.  I thought about just having a simple call in the middle to get a variable or constant at the top of the functions and have everything be the same... but felt that handicap some solvers... esp in function

```
constraint_complex_voltage_expansion_on_off{T <: AbstractWRForm}(pm::GenericPowerModel{T})
```
where a whole bunch of unnecessary constraints would be added that the solver (I am thinking those without good presolvers) might actually try to do something with... I looking for comments on whether this seems like a reasonable way to handle this...


ccoffrin:
Awesome!  I am going to review the code shortly, but two initial thoughts.

1) Regarding `constraint_complex_voltage`, the semantics of this constraints is, "all of constraints that link your voltage variables together"  In the case of a non-convex model, where there is no variable lifting, this does nothing.  In the case of a relaxations, which have lifting, this links the voltage variables together.  It's not always specific to the lines, for example in the QC relaxation this is where the constraint w_i = v_i^2 would go.  The SDP also has a similar property.

This `constraint_complex_voltage`, thing is one of the more counter intuitive aspects about the models right now, so I am open to ideas to improve/replace it.

2) The Matpower format was never intended to support EP.  I am thinking that we should extend it (and the corresponding json format) with the fields we need for this application, so that you can specify for example which corridors can be built and which cannot.  I think this will resolve many software engineering issues.

ccoffrin:
The code looks reasonable to me.  Some questions,

1) if you wanted to have a line that could be built, but is currently not in the network, how would you define that in the data file?

2) The formulation considered here is one where you have a 0/1 indicator for each line in the network.  One could also consider a version with {0,1,2,..,n} variable for the number of lines in a corridor.  How would we distinguish these two formulations?

src/prob/expansion.jl
```diff
@@ -0,0 +1,493 @@
...
+    end
+    
+    for (i,branch) in pm.set.branches
+        if !in(i,pm.data["new_branches"])    
+            constraint_active_ohms_yt_expansion(pm, branch)
+            constraint_reactive_ohms_yt_expansion(pm, branch)
+            constraint_phase_angle_difference_expansion(pm, branch)
+            constraint_thermal_limit_from(pm, branch)
+            constraint_thermal_limit_to(pm, branch)
+        end
```
rb004f: On 1) Yes, I was not aiming to be critical, but more wanted to start brainstorming how to handle that set of constraints.  It was the place I felt the least comfortable when implementing the expansion variables.

On 2) Yes, Matpower does not support EP.  If you pulled in a matpower case, this model would assume there is no expansion options unless a supplimental data file was provided.

On the comment "In these cases, why do you need to make new _expansion variants of the constraints, why not simply use the built in ones?".  Yes, I thought about this, but this would preclude creating a problem with expansion and switching options... one way to do this would be to split each of the on/off constraints into two functions.  One function takes as an argument the on/off variable and constructs the on/off constraint with this variable.  The second function(s) declares we will use the switching variable or the expansion variable and then calls the first function....

rb004f:
@ccoffrin I have updated the tnep data structures as we discussed. I release this fork to you.  A couple comments

1. I found a bug in calc_min_phase_angle of base.jl where the call to "sum" throws an exception if it is passed an array of size 1.  There is a fix for it in this pull request

2. There are some calculated constants in base,jl that can be incorrect if it does not have knowledge of tnep structures (for example calc_min_phase_angle).  This suggests that base.jl should be aware of tnep related data, the tnep problem will need to update some constants, or have its own  implementation of the function process_raw_data()

ccoffrin:
@rb004f I have reviewed and updated the code.  General points / thoughts,

- Thinking about other expansion problems we may have in the future, I am exploring the idea of having TNEP being a specific optimization problem and having NE (network expansion) be more general functions that could be re-used in other problem definitions.  One could imagine a CE (capacity expansion) problem, and a problem that does both.
- I made all of the new variables related to NE have an `_ne` postfix.  I think the code will be easier to debug if `p` and `p_ne` have different names.
- I generalized some of the helper functions in the core code so that those aspects don't need to be repeated.  This also prepared the code for a `calc_min_phase_angle` fix, although I have not applied that yet. 
- I removed some code that is only used in an the SDP relaxation.  I presume that is out of scope for now.
- It is worth to exploring how we could avoid repeating code across `_ne` line constraints and `_on_off` line constraints, however for now I suggest we leave that as a point in the issue tracker for later consideration.

ccoffrin:
Also, this code made it clear that this https://github.com/lanl-ansi/PowerModels.jl/issues/34, which is a good idea, which will simplify this code as well.

rb004f:
1. Agreed on being more general on expansion planning.  I considered putting in capacity expansion variables on this iteration, but need to get a couple other things done first.
2. Agreed on postfix change
3. Helper functions look good
4. Yeah, makes sense to keep SDP code out since it is not used yet
5. Yes, the repeated code across _ne and _on_off constraints bothers me... lets make it an issue and we can revisit.

Ready to merge into main branch?  Looks fine to me.

ccoffrin:
great. I'll fix the `calc_min_phase_angle` function and then merge in.

PR #36, opened 2016-10-26
TNEP function exports

rb004f:
This commit exports some TNEP functions so they can be used in other places...


ccoffrin:
Two thoughts,

1. Something is strange on this PR.  It includes a bunch of commits that were already merged.  I think the fork needs to be updated to the latest version of this repo.
2. The export command puts a symbol into the global scope and can clash with other exported functions, so this should be done with discretion.  You can call any function from inside a package with `PkgName.FunctionName` after `using PkgName`.

PR #38, opened 2016-11-09
adding support for sending raw json data to generic run method

ccoffrin:
title says it all.


ccoffrin:
Passes Linux tests, too lazy to wait for OS X.

PR #39, opened 2016-11-11
adding min requirements for compat and json

ccoffrin:


ccoffrin:
Works on linux builds.

tkelman:
Are you actually using features that did not exist before these versions? If not, these bounds are too tight.

ccoffrin:
As far as I know, we are not.  However, given that the package manger always pulls the most recent  feasible version, it is not clear to me how to compute the minimal required version of any particular package.  Suggestions?

I did check that this version of both of these packages is compatible with Julia v0.4,v0.5,v0.6.  I thought this would ensure a good amount of backward compatibility.  No?

tkelman:
If anyone pins or upper bounds one of these packages to something lower than thr bound you set here, they won't be able to install this package (or they'll be downgraded to versions that predate you bounding your dependencies). If they pin or upper bound to a really old version, then your package won't work because of missing features. Identifying the correct bounds depends on what features you're using from each package. Compat lists all the transformations it makes on its readme, and there's a note there showing how you can use blame on the readme to check which release first had a given feature. For JSON or other packages that don't list all their features in such a comprehensive way, you could check https://github.com/JuliaIO/JSON.jl/compare/v0.7.0...v0.8.0 and similar to see which commits / PRs are in each release. Or write a script to pin to different dependency versions and run your tests to see where they fail. We should look into automating this kind of thing.

ccoffrin:
I had a closer look, https://github.com/lanl-ansi/PowerModels.jl/pull/40.  Look resonable?

PR #40, opened 2016-11-11
relaxing requirements for compat and json

ccoffrin:


REQUIRE
```diff
@@ -1,5 +1,4 @@
 julia 0.4.6
-JSON 0.8.0
+JSON 0.6.0
 MathProgBase 0.5.4
 JuMP 0.14.0
-Compat 0.9.3
```
tkelman: can't remove Compat from require unless you also remove all uses of it

ccoffrin: As far as I can tell, this was never a req in the first place.  All test pass without including it.

tkelman: https://github.com/lanl-ansi/PowerModels.jl/blob/35ac71791b6dd86d3b6d55cf55c4d5cc2a573c54/src/PowerModels.jl#L8

the tests only pass because one of the particular versions of other dependencies that are getting used at the moment happen to depend on Compat - that's not guaranteed to always be the case

ccoffrin: Good catch.  I still have much to learn about this.

The core dependency in the code appears to be `readall -> readstring`, hence min version should be [compat v0.7.9](https://github.com/JuliaLang/Compat.jl/commit/8468afb0c11e1c51e2217913cc635b077f3064e6).

PR #53, opened 2016-12-04
Extensions to Matpower Format Parsing

### LINKED ISSUE #49: More Robust String Parsing in Matpower Files
Extend the split_line function in the Matpower parser to support splitting lines with multiple single-quoted string values.

ccoffrin:
This pull request addresses the need to get non-standard data into PowerModels.  It extends the Matpower parser to support a range of user extensions.

Documentation of these new features is provided in a new [DATA.md](https://github.com/lanl-ansi/PowerModels.jl/blob/mp-data-ext/DATA.md) file

This PR will close #49


ccoffrin:
Hey guys, I could use some input on this.  Both conceptual feedback and nitty gritty Julia implementation stuff.  There are some ugly hacks in there (e.g. see `build_typed_dict`, `split_line`), there should be a better way.

@rb004f @kersulis @kaarthiksundar @mlubin

rb004f:
The bigger question is how much matpower extension support do we want to provide? The json format is the standard, plus the ability to import other formats, like matpower... Unless there is a huge demand for extending Matpower, maybe the solution is to convert matpower to json, and then add extra data there?

ccoffrin:
@rb004f, this is a good point.  So far, I have had requests from @kaarthiksundar, @bluejuniper, and @MowenLu to be able to get extra data into PowerModels from an extended Matpower file.  Given our current user base, that feels like a "huge demand" to me.  This was my motivation for adding this feature.

Recently, I have also been observing that the raw text of large JSON files is not easy for humans to read, comprehend, or edit.  One needs a JSON viewer and/or a custom viewer, like the one I developed for the GRG JSON format, to understand the data.  In contrast, it is easy for people to read and debug small-ish matrixes in the Matpower format.

Also, I would not want to rally the community around PowerModel's current JSON format.  For me its really a internal thing, which can also be used for data exchange between algorithms.  If we are going to have a new network data standard, I would prefer something more wholistic, like GRG.  For now, I see Matpower as the standard data exchange format for research, hence we should have good support for it.  In time, I hope we can transition to something better, but until then, I think we should have good support for Matpower data.

PR #57, opened 2017-01-13
Updating syntax for Julia v0.5 and JuMP v0.15

ccoffrin:


ccoffrin:
Need to wait for next version of Pajarito before tests will pass. @mlubin

Issue #34, opened 2016-10-16
PowerDataSets to Dict{Symbol,Any}


ccoffrin:
To make GenericPowerModel more extensible.

Possibly remove ext.

PR #58, opened 2017-01-16
Reworking Data Processing

ccoffrin:
- Refactored Matpower data processing to steamline the internal network data structure and simplify editing data after parsing
- Added documentation on internal JSON data format to DATA.md
- Replaced PowerModels set data types with "ref" dictionary (see #34)
- Replaced JSON test files with Matpower test files
- Updated TNEP models to work with Matpower parsing extensions

These updates were motivated by an observation that a number of users are using PowerModels just for the Matpower parser and a number of inquiries on how to modify test cases in Julia.


ccoffrin:
@rb004f, these changes had significant impacts the internal JSON representation.  The previous .json files are not compatible with this version, hence why I moved the test cases to the .m format.

There was also a significant impact on the TNEP problem.  I renamed the TNEP branch matrix from `branch_ne` to `ne_branch`, so that it would play nice with the Matpower parsing extensions developed in #53.  Have a look at the new `*_tnep.m` test files.   Internally, the data associated with this matrix is available in `pm.data["ne_branch"]` (per the generic Matpower extension feature) and the pre-computed sets are in the new reference dictionary, `pm.ref[:ne_<item name>]`.

PR #69, opened 2017-01-24
More Robust Support for Modify Network Data in Julia

ccoffrin:
- made sure that computation of derived values is run before any solve.
- added tests for modifying network data in Julia
- minor improvements to Matpower parser
- Strengthened convex models with Lifted Nonlinear Cuts
- added tests for code snippets in documentation
- minor reorganization of variable and constraint functions


ccoffrin:
@rb004f, I think the only thing that may effect you is the calling of `update_derived_values`.  After some thought I decided to focus on making the default behavior possibly slow but always correct rather than fast and possibly incorrect.  Now `update_derived_values` is run on every call of `run_generic_model`.  As long as your code calls `run_generic_model` in some way, you don't need to think about `update_derived_values`.

PR #72, opened 2017-01-29
Constraint Templates

ccoffrin:
Added the idea of Constraint Templates to provide an abstraction layer between the network data and network constraint definitions.  This combined with some other data tiding eliminated the need for any "derived values" in the data format.

Also added a way to easily inspect the JuMP model produced by PowerModels.


ccoffrin:
@rb004f merging this in to move on to the next refactoring.  I don't think these changes will effect dependent codes.

PR #74, opened 2017-01-29
Refactoring Problem Formulations to use Complex Numbers

ccoffrin:
Previously constructors and constraints had annotations like "active", "reactive", and "complex" to indicate the space they operated on.  In this PR I am experimenting with making all constructors and constraints implicitly "complex", with a few explicitly calling out "active"/"reactive" if they only apply to a specific subset.

I think this approach will help the code match the math.  For example, it is nearly one-to-one with the formulations presented [here](https://arxiv.org/abs/1502.07847).

I am not 100% confident this is the best way to go, will wait a few days to merge in.  Discussion is welcome. @rb004f @kersulis @kaarthiksundar @mlubin @bluejuniper


ccoffrin:
Had discussions with a few LANL folks, all agreed this is a good idea.

PR #78, opened 2017-02-02
Removed Arrays from Internal Network Data Format

ccoffrin:
The arrays in the internal network format are replaced by dictionaries with the component "index" as the key value.  This enables sparse updates to the network data.

Other minor updates
- reworked how per-unit conversion is done
- make solution building utilize per-unit/mixed-unit conversion tools
- generalized some data wrangling code so it can be applied to both Matpower and JSON documents

### DIFF
--- a/src/io/common.jl
+++ b/src/io/common.jl
@@ -1,8 +1,18 @@
 function parse_file(file)
     if endswith(file, ".m")
-        return PowerModels.parse_matpower(file)
+        pm_data = PowerModels.parse_matpower(file)
     else
-        network_data = PowerModels.parse_json(file)
-        return network_data
+        pm_data = PowerModels.parse_json(file)
     end
+
+    check_network_data(pm_data)
+
+    return pm_data
 end
+
+function check_network_data(data::Dict{AbstractString,Any})
+    make_per_unit(data)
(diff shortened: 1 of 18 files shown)


ccoffrin:
@rb004f @kaarthiksundar, FYI changes to internal data format as discussed.  Using this new structure, `PowerModels.update_data(data, data_detla)` can be used to provide a sparse update to the network data.

src/io/common.jl
```diff
@@ -1,8 +1,18 @@
...
+        pm_data = PowerModels.parse_matpower(file)
     else
-        network_data = PowerModels.parse_json(file)
-        return network_data
+        pm_data = PowerModels.parse_json(file)
     end
-end
\ No newline at end of file
+
+    check_network_data(pm_data)
```
yeesian: should `check_network_data` happen inside `parse_matpower`, `parse_json`, etc? It modifies `pm_data` (see my comments in #95), and it feels weird to have `parse_matpower`, `parse_json`, etc return something different from `parse_file`.

ccoffrin: There are some special cases when one would not want to run `check_network_data`, but in the novice user case, it should always be run by default, if the provide a a file name or raw string data.

PR #86, opened 2017-03-08
Fix Bus Name Merge Bug

ccoffrin:
Fixes an issue that prevented parsing of bus_names when buses had non-contiguous ids.


ccoffrin:
Linux builds passed, so merging.

PR #96, opened 2017-03-31
Update for julia0.6 type system

yeesian:


ccoffrin:
Looks reasonable.  I presume `@compat` is for backward compatibility with v0.5?  If so, we probably need to update the minimum version of compat in REQUIRE.

yeesian:
> I presume @compat is for backward compatibility with v0.5? If so, we probably need to update the minimum version of compat in REQUIRE.

Yeah it is. I haven't checked which version of Compat that is.

ccoffrin:
Tested on my end.  This is looking good.  Thanks!

PR #103, opened 2017-04-04
Add Developer Documentation

### LINKED ISSUE #13: Documentation - Developers Guide
Add a special documentation guide / tutorial for developing new variable formulations and problem specifications.

yeesian:
Fixes #13.

### DIFF
--- a/src/core/base.jl
+++ b/src/core/base.jl
@@ -27,24 +27,7 @@ where
 * `setting` usually looks something like `Dict("output" => Dict("line_flows" => true))`, and
 * `ref` is a place to store commonly used pre-computed data from of the data dictionary,
     primarily for converting data-types, filtering out deactivated components, and storing
-    system-wide values that need to be computed globally. Some of the common keys include:
-
-    * `:off_angmin` and `:off_angmax` (see `calc_theta_delta_bounds(data)`),
-    * `:bus` -- the set `{(i, bus) in ref[:bus] : bus["bus_type"] != 4}`,
-    * `:gen` -- the set `{(i, gen) in ref[:gen] : gen["gen_status"] == 1 && gen["gen_bus"] in keys(ref[:bus])}`,
-    * `:branch` -- the set of branches that are active in the network (based on the component status values),
-    * `:arcs_from` -- the set `[(i,b["f_bus"],b["t_bus"]) for (i,b) in ref[:branch]]`,
-    * `:arcs_to` -- the set `[(i,b["t_bus"],b["f_bus"]) for (i,b) in ref[:branch]]`,
-    * `:arcs` -- the set of arcs from both `arcs_from` and `arcs_to`,
-    * `:bus_arcs` -- the mapping `Dict(i => [(l,i,j) for (l,i,j) in ref[:arcs]])`,
-    * `:buspairs` -- (see `buspair_parameters(ref[:arcs_from], ref[:branch], ref[:bus])`), and
-    * `:bus_gens` -- the mapping `Dict(i => [gen["gen_bus"] for (i,gen) in ref[:gen]])`.
-
-    If `:ne_branch` exists, then the following keys are also available with similar semantics:
-
-    * `:ne_branch`, `:ne_arcs_from`, `:ne_arcs_to`, `:ne_arcs`, `:ne_bus_arcs`, `:ne_buspairs`.
-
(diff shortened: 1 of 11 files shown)


src/core/base.jl
```diff
@@ -6,16 +6,58 @@ export
...
+    setting::Dict{String,Any}
+    solution::Dict{String,Any}
+    ref::Dict{Symbol,Any} # reference data
+end
+```
+where
+
+* `data` is the original data, usually from reading in a `.json` or `.m` (patpower) file
+* `setting` usually looks something like `Dict("output" => Dict("line_flows" => true))`, and
+* `ref` is a `Dict` with the following keys:
```
ccoffrin: I think this level of detail about ref, should be left to the function that populates the ref data structure.  Here is an overview of what this is for,

Ref is a place to store commonly used pre-computed data from of the data dictionary.  The three primary uses are,
1) convert dictionary lookups from "string" => object to int => object, so that the type of ids inside objects is the same as the lookup.
2) filter out components that are deactivated in the data
3) store some system-wide values that need to be computed globally.  Such as identifying the reference bus or computing `off_angmin`.

src/core/base.jl
```diff
@@ -6,16 +6,58 @@ export
...
+
+    * `:off_angmin` and `:off_angmax` (see `calc_theta_delta_bounds(data)`),
+    * `:bus` is the set `{(i, bus) in ref[:bus] : bus["bus_type"] != 4}`,
+    * `:gen` is the set `{(i, gen) in ref[:gen] : gen["gen_status"] == 1 && gen["gen_bus"] in keys(ref[:bus])}`,
+    * `:branch` is the set `{(i, branch) in ref[:branch]` satisfying the conditions
+
```
ccoffrin: more simply put, this is the set of branches that are active in the network, based on the component status values.

src/form/acp.jl
```diff
@@ -76,7 +86,14 @@ function constraint_kcl_shunt_ne{T <: AbstractACPForm}(pm::GenericPowerModel{T},
...
 
-"Creates Ohms constraints (yt post fix indicates that Y and T values are in rectangular form)"
+"""
+Creates Ohms constraints (yt post fix indicates that Y and T values are in rectangular form)
+
+```
+p[f_idx] == g/tm*v[f_bus]^2 + (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) + (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))
+q[f_idx] == -(b+c/2)/tm*v[f_bus]^2 - (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) + (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))
+```
+"""
```
ccoffrin: Is it possible that the docs can support something like MathJax for latex type setting?  If so, we can write some very clean version of these constraints.

yeesian: I think we can use `doc""` to get latex behaviour. We can also do that in a future PR.

src/core/base.jl
yeesian: If you prefer, I can change it to move the current description of `ref` to `build_ref()`, and use your comment in place of the description in `GenericPowerModel`.

It appears to me that `GenericPowerModel` is the exported "public-facing" interface, and is likely to be where users look up information on how to use `ref`. The docs were written with the consideration of letting them know the common keys that will be available, rather than having to be re-directed to look up the documentation of an internal function.

Anyhow, let me know what you decide? I'll make the changes accordingly.

ccoffrin: > If you prefer, I can change it to move the current description of ref to build_ref(), and use your comment in place of the description in GenericPowerModel.

I think this is the best plan for now.  An in-depth explanation of how to utilize ref to build a model is best left  for an in-depth tutorial.  It's really only there as a convenience, users can work from the data directly, if they wish.

yeesian: See https://github.com/lanl-ansi/PowerModels.jl/pull/103/commits/280405527eada0aacb288a81012b8bb9e6d477c0

ccoffrin:
Looks good.  Thanks!

PR #107, opened 2017-06-03
Adding Some Basic OPF Results to Docs

ccoffrin:
@yeesian, planning on adding some basic results documentation. Suggestions welcome.


yeesian:
👍 

I think a basic example complete with a script, or documentation for what to call to get the results will be good.

ccoffrin:
@yeesian, good suggestion.  I didn't want to get overly detailed, but I think what I have added gives the key points on how to build the table.  Look good to you?

yeesian:
Yeah it looks much better now.

---

I might add a quick example like

```julia
> using PowerModels, Ipopt
> result = run_ac_opf("$(Pkg.dir("PowerModels"))/test/data/case30.m", IpoptSolver())
```

should yield a dictionary like

```julia
Dict{String,Any} with 8 entries:
  "solve_time"   => 7.34359
  "status"       => :LocalOptimal
  "objective"    => 204.968
  "solver"       => "Ipopt.IpoptSolver"
  "solution"     => Dict{String,Any}(Pair{String,Any}("baseMVA",100.0),Pair{String,Any}("gen",Dict{String,Any}(Pair{String,Any}("4",Dict{String,Any}(Pair{String,Any}("qg",40.0…
  "data"         => Dict{String,Any}(Pair{String,Any}("name","nesta_case30_ieee"),Pair{String,Any}("bus_count",30),Pair{String,Any}("branch_count",41))
  "machine"      => Dict("cpu"=>"Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz","memory"=>"16.0 Gb")
  "objective_lb" => -Inf
```

which you can then query for relevant information to put into the table, e.g.
```julia
julia> result["data"]
Dict{String,Any} with 3 entries:
  "name"         => "nesta_case30_ieee"
  "bus_count"    => 30
  "branch_count" => 41
```

to get the values corresponding to `Case Name`, `Nodes`, and `Edges` respectively.

ccoffrin:
Ok great.  As for this detailed example you suggest, I think this would be better placed as an example in the documentation of the "results" data structure.  This area of the documentation should focus on the experiment and assume the reader is fairly experienced with PowerModels.

yeesian:
I have no qualms with that when there is documentation of the "results" data structure that readers can be pointed to.

PR #129, opened 2017-07-21
Adding Deprecated constraint_theta_ref Function

ccoffrin:
Adds a depreciation warning for this function, so that this change is non-breaking in next version.

### DIFF
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -10,6 +10,7 @@
 # and should never refer to model variables
 #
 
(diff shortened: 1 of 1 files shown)


ccoffrin:
@mlubin could not figure out how to best deprecate a function with a different signature, any tips?

mlubin:
``@Base.deprecate constraint_theta_ref(pm::GenericPowerModel) ...`` should deprecate only the matching signiture

Issue #112, opened 2017-06-28
Support towards multiperiod and/or stochastic OPF 


frederikgeth:
A few colleagues and I are considering contributing to PowerModels.jl to support multiperiod and/or stochastic OPF. We streamed your JuMP developers meetup presentation.

You'd need (at least) one additional dimension for each variable in your (optimal) power flow model to represent time (multiperiod) or the decision stages & samples (multistage stochastic programming). 

How would you attack this? Given the current implementation you'd need to add this dimension to each formulation in /form, or is there some JuMP syntax to avoid this?

Do you see unit-commitment type models (load, generation, storage) being part of PowerModels or separated into a separate project eventually? Just like the gas grid flow models are already? 

Obviously such functionality would go far beyond the scope of the .m matpower case format, so how would you want to evolve PowerModels' internal datamodel and the transformations to/from .m and json? Or other dataformats?

ccoffrin:
Hi @frederikgeth thanks for your interest in power models!

In terms of scope, for sure Stochastic OPF, Secure Constrained OPF, and UC OPF are in the scope of the PowerModels package.

My current idea to support problems like Secure Constrained OPF and UC OPF is to add the notion of network replicates, see https://github.com/lanl-ansi/PowerModels.jl/issues/44.  As you suggest, all variable indexes would have have a new _replicate_ dimension (which is 0 by default).  Then, when one is writing a model, they can make multiple copies of the network and link the variables together with additional constraints.  This is not the absolute best encoding of such problems, but I think it's a good balance of implementation flexibility and simplicity.

As for datasets, have you seen that PowerModels supports [extensions to the Matpower format](https://lanl-ansi.github.io/PowerModels.jl/stable/data.html)? Ideally we would be able to define some standardized extensions to the Matpower format so that they are cross-compatible with PowerModels and Matpower.  Very large time series data can be prohibitive to this approach; but I would like to start by keeping the formulation and data management as simple as possible.

Another open challenge with these problem domains is curating a reasonable collection of test cases from the literature.  I am coordinating this [IEEE PES Task Force](https://power-grid-lib.github.io/) to provide a common location for such collections of test cases.

I'll also bring your inquiry to the attention of @yeesian and @kaarthiksundar, who have spoken to me about developing similar extensions.

yeesian:
Kaarthik and I had indeed spoken about developing similar extensions, but don't yet have anything concrete. I'll just mention some of the things we've spoken about:

---

[...]

Issue #44, opened 2016-11-16
Consider adding a replicate index to all variables


ccoffrin:
To more readily support multi-time models.

ccoffrin:
from @rb004f, "More generally we'd also want to index by scenario, or perhaps other reasons"

Good point!  It should be a fairly general way to include multiple network copies in one JuMP model.  The constraints you add to link those variables together would differ based on these different applications.

ccoffrin:
To enable this we need to, move all variable constructors to [JuMP anonymous variables](http://www.juliaopt.org/JuMP.jl/latest/refvariable.html#Anonymous-variables-1) and add our own methods for organizing and looking up these variables.  This will also will resolve the issue of how to have two Infrastructure Models in one JuMP model.

ccoffrin:
Proposal, `pm.var[:p]` for looking up variables.

PR #150, opened 2017-08-06
Idea for Supporting Multiple Networks

ccoffrin:
@yeesian @rb004f @kaarthiksundar @frederikgeth @hakanergun

This is what I had in mind for how PowerModels could support multiple networks in one JuMP model, as discussed in #44, #112.  This is a work in progress, currently only the OPF problem is updated, but I think this sufficient to have a useful discussion on the design.

The core idea is to introduce a notion of "networks"(nw) into the data object, this same organization then follows through to all of the other structures in GenericPowerModel, i.e. ref, var, ext.

The top-level data schema becomes,
```
{
  "nw":{
    {"1":{<current single network data>},
    {"2":{<current single network data>},
    ...
  }
}
```
Problem metadata that spans multiple networks can be put at the top level of this data object.  Things like ref, var, ext have the following top-level organization,
```
{
  :nw:{
    {1:{<current single network items>}},
    {2:{<current single network items>}},
    ...
  }
}
```
Which follows the same organizational convention of components within in the networks.  Currently when we read in a single network data file (i.e. Matpower), it is assign the default network index of "0".

I have mocked up an example Multi-Period OPF (mpopf) problem to show how I imagined this structure would be used to do multi-network optimization,
https://github.com/lanl-ansi/PowerModels.jl/blob/networks/src/prob/mpopf.jl
https://github.com/lanl-ansi/PowerModels.jl/blob/networks/test/mpopf.jl

I have also added some helper functions (e.g. nw_ref), to help modelers ignore the multi-network structure, if their problem is only one network,
https://github.com/lanl-ansi/PowerModels.jl/blob/networks/src/prob/opf.jl

Pros of this proposal (as is)
- the information structure is self consistent with the rest of PowerModels
- this design is extremely generic and flexible, for example you could load two entirely different networks into "nw" and it would work fine
- having an explicit "nw" dictionary allows for data and variables that span multiple networks
- Any single-network function is easily extended to this multi-network structure (e.g. `make_per_unit`)

Cons of this proposal (as is)
- if you are __not__ solving a multi-network problem there is complexity that will feel cumbersome.  For example getting data is now `data["nw"]["0"]["bus"]["1"]["vm"]`, this `["nw"]["0"]` seems tedious.
- the modeling looks complex and may be intimidating to developers.  For example  `pm.var[:nw][n][:w] = @variable(pm.model,  ... lowerbound = pm.ref[:nw][n][:bus][i]["vmin"]^2 ...)`
- the price we pay for extremely generality is that having many copies of a similar network can be a space hog.  That said, it should not add more than 2-3x the size of the optimization model it self.

I think some of these cons can be addressed through further restructuring of the software, but I wanted to get some initial feedback before getting into that level of refinement.

For those of you who have been interested in using such an extension.  I don't have a super clear idea of what specific problems you are interested in solving.  It would be a great help to me if you could mock-up what your problem would look like, i.e. your version of `post_mpopf`.  This will help me understand if this proposal will fit your needs and if it is worth pursuing further.

General comments/questions/suggestions are encouraged.

### DIFF
--- a/src/PowerModels.jl
+++ b/src/PowerModels.jl
@@ -30,6 +30,7 @@ include("form/shared.jl")
 
 include("prob/pf.jl")
 include("prob/opf.jl")
+include("prob/mpopf.jl")
 include("prob/ots.jl")
 include("prob/misc.jl")
 include("prob/tnep.jl")
--- a/src/core/base.jl
+++ b/src/core/base.jl
@@ -58,8 +58,13 @@ function GenericPowerModel(data::Dict{String,Any}, T::DataType; setting = Dict{S
     # i.e. https://github.com/lanl-ansi/PowerModels.jl/issues/131
 
     ref = build_ref(data)
-    vars = Dict{Symbol,Any}(n => Dict{Symbol,Any}() for n in keys(ref))
-    ext = Dict{Symbol,Any}(n => Dict{Symbol,Any}() for n in keys(ref))
+    
+    var = Dict{Symbol,Any}(:nw => Dict{Int,Any}())
+    ext = Dict{Symbol,Any}(:nw => Dict{Int,Any}())
+    for nw_id in keys(ref[:nw])
+        var[:nw][nw_id] = Dict{Symbol,Any}()
+        ext[:nw][nw_id] = Dict{Symbol,Any}()
+    end
 
     pm = GenericPowerModel{T}(
         Model(solver = solver), # model
@@ -67,7 +72,7 @@ function GenericPowerModel(data::Dict{String,Any}, T::DataType; setting = Dict{S
         setting, # setting
         Dict{String,Any}(), # solution
         ref,
-        vars,
+        var,
         ext
     )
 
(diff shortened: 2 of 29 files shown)


ccoffrin:
An idea for removing some of the verbosity in the network modeling.

Add the notion of a `current_network` to GenericPowerModel.  Then add functions like,
```
getref(pm, :bus)
getvar(pm, :bus)
getext(pm, :bus)
```
which translate into,
```
pm.ref[:nw][current_network][:bus]
```
set the default value of `n` in various methods to be `pm.current_network`

yeesian:
Sounds sensible. How do you see its relation to the replicates idea (c.f. https://github.com/lanl-ansi/PowerModels.jl/issues/112#issuecomment-318916942)?

ccoffrin:
Another idea is manage the verbosity of the models.  Add a field along the lines of `multinetwork:<bool>` the network data.  Then we can switch the function behavior based on this flag.

ccoffrin:
@kaarthiksundar suggested adding a helper function that transforms a single network into a multi-network with `n` deep copies of the original network, which can then be modified, for example with a load profile data.

frederikgeth:
The approach seems to be a good start for SCOPF as well.

Expected features of a multi-period OPF:
- network model just duplicated
- time steps with equal length, but preferably with support for time steps of different lengths.
- loading time series data for loads (and DG)
- modeling energy storage system dynamics (battery storage, pumped hydro, ...), 
- eventually including a cost model for storage as well (losses, wear).
- costs summed over time horizon

I can propose a formulation / contribute an implementation of such aspects; what timeline do you foresee?

I agree developing new formulations may be intimidating to new developers, but good documentation can go a long way.

Some helper functions for exploring results would indeed be nice. Furthermore, exploring multiperiod OPF results is hard even for somebody familiar with classical OPF. One solution is summarizing sets of time series using boxplots (voltage magnitude by node id, power by generator id, power flow by line id...) and bar plots (total energy generated by generator, costs of generator).

ccoffrin:
It seems like everyone is in favor of this approach so I will continue to hone it.  I expect this feature to be merged into master around August 28, but please feel free to explore modeling possibilities working off this branch.

ccoffrin:
This feature and discussion has moved to #166.

PR #151, opened 2017-08-16
Inclusion of shiftable PSTs and OLTCs

hakanergun:
The implementation of shiftable PSTs and OLTCs are ready from our side, also tested and validated.

Should we delete this pull request and send a pull request to PowerModelsAnnex or should it be part of PowerModels?


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ PowerModels.jl Change Log
 - Fixed bug in constants for w-space phase angle difference constraints
 - Fixed bug when no reference bus was specified
 - Fixed dcline parsing bug
+- Support for DC Line formulation compatible with matpower
```
ccoffrin: This is duplicate information.

README.md
```diff
@@ -45,6 +45,7 @@ The primary developer is Carleton Coffrin with support from the following contri
 - Miles Lubin (@mlubin) MIT, Julia/JuMP advise
 - Yeesian Ng (@yeesian) MIT, Documenter.jl setup
 
+Frederik Geth and Hakan Ergun (both KU Leuven) contributed to the Matpower-compatible DC Line formulation.
```
ccoffrin: also duplicate information

src/core/variable.jl
```diff
@@ -351,6 +351,40 @@ function variable_reactive_dcline_flow(pm::GenericPowerModel; bounded = true)
     return pm.var[:q_dc]
 end
 
+"variable: `t_shift[l,i,j]` for `(l,i,j)` in `arcs`"
+function variable_phase_shift(pm::GenericPowerModel; bounded = true)
```
ccoffrin: This class of variables does not need the bounded flag.  This is only necessary for variables that are used in Power Flow studies.

src/form/acp.jl
```diff
@@ -149,7 +149,53 @@ function constraint_ohms_yt_to{T <: AbstractACPForm}(pm::GenericPowerModel{T}, f
...
+
+"""
+Creates Ohms constraints for shiftable PSTs (yt post fix indicates that Y and T values are in rectangular form)
+
+```
+p[f_idx] == g*v_fr^2 + (-g)*(v_fr*v_to*cos((t_fr - t_shift_fr) - (t_to - t_shift_to))) + (-b)*(v_fr*v_to*sin((t_fr - t_shift_fr) - (t_to - t_shift_to)))
+q[f_idx] == -(b+c/2)*v_fr^2 - (-b)*(v_fr*v_to*cos((t_fr - t_shift_fr) - (t_to - t_shift_to))) + (-g)*(v_fr*v_to*sin((t_fr - t_shift_fr) - (t_to - t_shift_to)))
+```
+"""
+function constraint_ohms_yt_from_shiftable{T <: AbstractACPForm}(pm::GenericPowerModel{T}, f_bus, t_bus, f_idx, t_idx, g, b, c)
```
ccoffrin: The name of this constraint should be along the lines of `constraint_ohms_y_from_shiftable`.  The `yt` block indicates which constants are represented in rectangular coordinates.  In this case only the admittance is rectangular.  

This equation is also missing the tap ratio term, which I believe is easy to add.  Have a look at the `constraint_ohms_y_from` constraint for an example.

src/form/acp.jl
```diff
@@ -149,7 +149,53 @@ function constraint_ohms_yt_to{T <: AbstractACPForm}(pm::GenericPowerModel{T}, f
...
+    p_fr = pm.var[:p][f_idx]
+    q_fr = pm.var[:q][f_idx]
+    v_fr = pm.var[:v][f_bus]
+    v_to = pm.var[:v][t_bus]
+    t_fr = pm.var[:t][f_bus]
+    t_to = pm.var[:t][t_bus]
+    t_shift_fr = pm.var[:t_shift][f_idx]
+    t_shift_to = pm.var[:t_shift][t_idx]
+
+    c1 = @NLconstraint(pm.model, p_fr == g*v_fr^2 + (-g)*(v_fr*v_to*cos((t_fr - t_shift_fr) - (t_to - t_shift_to))) + (-b)*(v_fr*v_to*sin((t_fr - t_shift_fr) - (t_to - t_shift_to))))
```
ccoffrin: It is not clear to me why we have two shift variables, `t_shift_fr` and `t_shift_to`.  The model I have seen has one shift variable and takes a form along these lines,
```
p_fr = ... cos(t_fr - t_to - shift) ...
p_to = ... cos(t_to - t_fr + shift)
```

test/data/case5_dc.m
```diff
@@ -1,8 +1,28 @@
+<<<<<<< HEAD
+% NESTA v0.6.0
+=======
```
ccoffrin: These merge conflicts need to be removed.  😄

test/data/case5_pst.json
```diff
@@ -0,0 +1,397 @@
+{
```
ccoffrin: In the interest of keeping maintenance simple, the PowerModels test data are all kept in the Matpower format.  You can use the generic [Matpower extensions feature](https://lanl-ansi.github.io/PowerModels.jl/stable/network-data.html) to get the additional data you need into the PowerModels data dictionary.

src/form/acp.jl
frederikgeth: we're following [this](https://pdfs.semanticscholar.org/bdd9/d4b3b0cc625ad81967594398f2a7119d210f.pdf) reference for the line model (Fig. 2.9) 

in any case, for voltage magnitude transformation, you may have a static (off-load) tap on one side and an on-load tap changer on the other side. So this part requires to and from variables. And it feels a bit weird not to let the line model be fully symmetric.

ccoffrin: I see.  My concern is then that the optimization model formulation should match the data format.  The Matpower assumes a simpler pi-model.  So, if you want to develop a network model based on Fig. 2.9 you should also extend the data format to explicitly model this other type of pi-model.

For what its worth, in the past I have had success building models with OLTP by making Matpower's `ratio` and `shift` parameters decision variables (e.g. see, http://sites.ieee.org/psace-mho/panels-and-competitions-2014-opf-problems/).  I cannot immediately see the added value of the "transformer symmetrical model", but this is byond my knowledge of Power Engineering.  Maybe you can provide some motivation for this model over the Matpower one?

test/pf.jl
```diff
@@ -18,6 +18,24 @@
         @test isapprox(result["solution"]["dcline"]["1"]["qf"], -0.403045; atol = 1e-5)
         @test isapprox(result["solution"]["dcline"]["1"]["qt"],  0.0647562; atol = 1e-5)
     end
+    @testset "3-bus case with active DC line" begin
```
frederikgeth: weird, @hakanergun we have some more weird merge issues to fix.

src/form/acp.jl
frederikgeth: It matters on which side you physically put the voltage magnitude tap changer (easy to see that losses are impacted). You see both options being modeled also [elsewhere](https://nl.mathworks.com/help/physmod/sps/examples/oltc-regulating-transformer-phasor-model.html). Typically, an OLTC will be on the 'from' side - the OLTC device will have to handle much less current. We could implement a 'from'-only formulation here and add the extensions to PowerModelsAnnex.jl. 

For a PST it's a matter of signs (shift(to-> from) = -shift(from->to)). But it feels weird having asymmetric PST parameters if the OLTC part is symmetric. 

In any case, I'm inclined to rename the default fields `tap` to `tap_fr` and `shift` to `shift_fr`, to be more clear w.r.t. what the matpower parameters actually _represent_ (cfr. [Fig 3-1](http://www.pserc.cornell.edu/matpower/MATPOWER-manual.pdf)). 

Furthermore, we aim to add active power losses (no-load) in the transformer model as well (e.g. as shunt conductance of the transformer), to satisfy modelling requirements for distribution systems (such losses are not equally negligible in distribution as in transmission). Which direction would you want to go here:

- add conductance to line equations in general (g_shunt = 0 for backwards compatibility)
- separate transformer models from normal line models with support for shunt conductance

I feel that extensions should include the matpower line model at least as an edge case (for backwards compatibility). But sticking to a simple pi model implies certain limitations (i.e. equally focussed on transmission system modeling accuracy tradeoffs). What's your take on this?

hakanergun:
The implementation of shiftable PSTs and OLTCs are ready from our side, also tested and validated.

Should we delete this pull request and send a pull request to PowerModelsAnnex or should it be part of PowerModels?

src/core/variable.jl
```diff
@@ -299,6 +299,60 @@ function variable_reactive_dcline_flow(pm::GenericPowerModel; bounded = true)
...
+    for (l,i,j) in pm.ref[:arcs_from]
+        shift_min[(l,i,j)] = pm.ref[:branch][l]["shift_fr_min"]
+        shift_max[(l,i,j)] = pm.ref[:branch][l]["shift_fr_max"]
+        shift[(l,i,j)] = pm.ref[:branch][l]["shift_fr"]
+    end
+    for (l,i,j) in pm.ref[:arcs_to]
+        shift_min[(l,i,j)] = pm.ref[:branch][l]["shift_to_min"]
+        shift_max[(l,i,j)] = pm.ref[:branch][l]["shift_to_max"]
+        shift[(l,i,j)] = pm.ref[:branch][l]["shift_to"]
+    end
```
ccoffrin: If these computations are used some other place, you may consider adding them to` pm.ext`

src/io/matpower.jl
```diff
@@ -470,7 +479,11 @@ function parse_matpower_data(data_string::String)
                     "rate_b" => parse(Float64, branch_row[7]),
                     "rate_c" => parse(Float64, branch_row[8]),
                     "tap" => parse(Float64, branch_row[9]),
+                    "tap_fr" => parse(Float64, branch_row[9]),
+                    "tap_to" => 1,
                     "shift" => parse(Float64, branch_row[10]),
+                    "shift_fr" => parse(Float64, branch_row[10]),
+                    "shift_to" => 0,
```
ccoffrin: Same as above.

src/io/matpower.jl
```diff
@@ -49,9 +49,18 @@ function update_branch_transformer_settings(data::Dict{String,Any})
...
+                if branch["tappable"] == 1 || branch["shiftable"] == 1
+                    branch["transformer"] = true
+                else
+                    branch["transformer"] = false
+                end
+            else
+                branch["transformer"] = false
+            end
             branch["tap"] = 1.0
+            branch["tap_fr"] = 1.0
```
ccoffrin: Because this formulation is outside the Matpower spec, these kinds of data updates should happen outside of the standard Matpower workflow.

src/prob/tfopf.jl
```diff
@@ -0,0 +1,55 @@
...
+end
+
+""
+function post_tfopf(pm::GenericPowerModel)
+    variable_voltage(pm)
+    variable_generation(pm)
+    variable_line_flow(pm)
+    variable_dcline_flow(pm)
+    variable_phase_shift(pm)
+    variable_voltage_tap(pm)
```
ccoffrin: The current convention in power models is that variables and constraints are based on complex numbers.  So in this case, the variable and constraint signature should be along the line of "variable_transformation" or "variable_oltp".  Then inside "variable_transformation" you call "variable_voltage_shift" and "variable_voltage_tap".  For example see, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/variable.jl#L182

src/prob/tfopf.jl
```diff
@@ -0,0 +1,55 @@
...
+
+    for (i,bus) in pm.ref[:bus]
+        constraint_kcl_shunt(pm, bus)
+    end
+
+    for (i,branch) in pm.ref[:branch]
+        if branch["shiftable"] == false && branch["tappable"] == false
+            constraint_ohms_yt_from(pm, branch)
+            constraint_ohms_yt_to(pm, branch)
+            constraint_link_voltage_magnitudes(pm, branch)
```
ccoffrin: seems like this constraint should be in the else clause?

test/tpopf.jl
```diff
@@ -0,0 +1,32 @@
+
+
+@testset "test ac polar tfopf" begin
+    @testset "5-bus tf case" begin
+        result = run_ac_tfopf("../test/data/case5_tf.m", ipopt_solver)
+
+        @test result["status"] == :LocalOptimal
+        @test isapprox(result["objective"], 15155.17; atol = 1e0)
+        @test isapprox(result["branch"]["shiftf"], XXX; atol = 1e0)
+        @test isapprox(result["branch"]["tapf"], XXX; atol = 1e0)
```
ccoffrin: Should these XXX have values?

Issue #20, opened 2016-08-30
Give constraints names and return dicts of symbols 


ccoffrin:
similar to variables.

ccoffrin:
closed by #158

PR #158, opened 2017-08-23
Remove Variable and Constraint Return Values

### LINKED ISSUE #124: Use in-place Union of Constraint Ids
When collecting constraint ids use `union!` instead of `union`, the later is problematic for very large cases.

ccoffrin:
These were not tested and partly broken and also causing a significant performance issue, so dropping support for this feature seems apt.  Closes #124.


ccoffrin:
Also makes #20 irrelevant.

PR #166, opened 2017-08-27
Support for Multi-Networks

ccoffrin:
@yeesian @rb004f @kaarthiksundar @frederikgeth @hakanergun

There were so many updates on master I restarted #150 from scratch and here it is, complete support for multi-networks! 🎉 I need to wait to merge this into master until at least v0.4.0 has made its way into METADATA.jl, https://github.com/JuliaLang/METADATA.jl/pull/10931.  So we have a few days for review and discussion.

Overall I believe I have incorporated all of the ideas discussed in #44, #112, and #150.  I am fairly happy with the result, but still open to suggestions / feature requests.  A big benefit of this current implementation is that only expert users should need to understand the multi-network abstractions, I hope.  

Here are a few big changes that should be noted,

### Network Data
This PR supports two types of network data.  The original network data now has the form,
```
{
  "multinetwork":false,
   <current single network data>
}
```
and the new network data has the form,
```
{
  "multinetwork":true,
  "nw":{
    {"1":{<current single network data>},
    {"2":{<current single network data>},
    ...
  }
}
```
The same structure holds true for the solutions that come out of PowerModels.

### Ref and Var
To make the implementation simpler for developers, under the hood GenericPowerModel only has one information organization, which is a the more general multi-network design.  For example,
```
pm.ref[:nw][n][:gen]
pm.var[:nw][n][:pg]
```
GenericPowerModel also has a new field, current network `cnw`.  This is used as the default network id `n`, when a network id is not provided explicitly.  Helper functions leverage `cnw` so that a novice user does not need to know about multinetworks.  For example,
```
pm.ref[:nw][pm.cnw][:gen] <=> ref(pm, :gen)
pm.var[:nw][pm.cnw][:pg][i] <=> var(pm, :pg)[i] or var(pm, :pg, i)
```
The current convention is that user facing code uses the helper functions `ref(pm,...), var(pm,...), ids(pm,...)` and backend code uses the explicit form `pm.ref[:nw][pm.cnw]...`.

### Problem Specifications
Previously problem specifications had a structure like this,
```
for (i,bus) in pm.ref[:bus]
   constraint_kcl_shunt(pm, bus)
end
```
In the presence of multiple networks and the `cnw` value, this structure became error prone.  So it has changed to,
```
for i in ids(pm, :bus)
   constraint_kcl_shunt(pm, i)
end
```
It is now the job of the `constraint_kcl_shunt` constraint template to look up the correct bus data for bus `i` in the correct `cnw` network data.

### Multi-Network Modeling

This is very similar to what was discussed in #150 and now lives [here](https://github.com/lanl-ansi/PowerModels.jl/blob/nws/test/multinetwork.jl).

As always, comments/questions/suggestions are encouraged.


rb004f:
The alternative way to support multi-network is to read in multiple json files, 1 per network. This is currently how I handle PowerModels.jl + GasModels.jl models, mainly so I don't have to add a label about what type of network each one is.  I think we want to be consistent in how we handle multiple intra-networks and multiple inter-networks. I can be convinced of either approach, but we should think about the pros and cons of each

ccoffrin:
Given what we have here, I would probably organize a multiple types of infrastructure as,
```
{
  "inf"{
    "pg":{
       <package specific network spec.>
    }
    "ng":{
      <package specific network spec.>
    }
  }
}
```
But we don't need to decided this immediately to move forward with this feature in PowerModels.

rb004f:
Its more a data representation philosophy question.  I think the choice of philosophy in PowerModels should be adopted for representing multiple networks.  One uber file or multi files?  As I said, I am fine to go either direction.

ccoffrin:
At this point I don't have a good feeling for the best software design for integrating two such packages.  Typically I just try to implement it and see what design decisions cause pain and revise things as needed.  If that causes a significant revision of how this is currently done, that's fine by me.

At this point adding support for multiple power networks has added a lot of new abstraction to this code.  I recommend that we get our heads around this change and make sure we are happy with it, before abstracting further over multiple types of infrastructures.

ccoffrin:
@yeesian @rb004f @kaarthiksundar @frederikgeth @hakanergun last call for comments on this before I merge into master.

frederikgeth:
I don't have any objections to the implementation (but I didn't have much time to explore). I think we could easily build upon it for SCOPF and multi-period OPF. The entry for new contributors is a bit higher now, but given some documentation it shouldn't be an issue.

lroald:
@ccoffrin 
I was trying to find the documention/example for multi-networks (which I distinctly remember from before), but the ling appears to be broken. Are there any examples available that we can look at that uses multi-networks?

ccoffrin:
Good question.  This is not yet well documented!  Looks like I should prioritize this issue, https://github.com/lanl-ansi/PowerModels.jl/issues/169

In the meantime have a look at these examples,

Problem Specifications:
https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/prob/test.jl#L73
https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/prob/test.jl#L306

Setting Up Multinetwork data:
https://github.com/lanl-ansi/PowerModels.jl/blob/master/test/common.jl

lroald:
Thanks! This already gives a much better starting point for understanding!

yeesian:
I just realized I forgot to include [notes on the data schema](https://github.com/lanl-ansi/PowerModels.jl/pull/150#issue-134332656), but Here's a first shot at documenting it: https://lanl-ansi.github.io/PowerModels.jl/latest/multi-networks/. Sorry it took me so long!

PR #183, opened 2017-10-23
Support for Dual Values in Solutions

ccoffrin:
A preliminary implementation of how to add dual values to PowerModels solutions.

Usage example:
```
result = run_dc_opf("case.m", IpoptSolver(), setting = Dict("output" => Dict("duals" => true)))
PowerModels.make_mixed_units(result["solution"])
result["solution"]["bus"]
```
A key open question is what is the best way to initialize the constraint reference dictionaries structures.  Constraint temples work, but is this the best place?


ccoffrin:
Before merging this branch requires unit tests.

ccoffrin:
Integrated in #194

PR #191, opened 2017-11-03
qc trilinear convex hull

kaarthiksundar:


ccoffrin:
Code looks good.  Make sure to add a test and a note to the change log.  The case `case5_asym.m` is a good one to test because it has nontrivial bounds on the voltage angle difference.

src/form/wr.jl
```diff
@@ -736,3 +737,97 @@ function constraint_power_magnitude_link_on_off(pm::QCWRPowerModel, n::Int, i, a
...
+        if buspair["branch"] == i
+            constraint_power_magnitude_sqr(pm, n, i)
+            constraint_power_magnitude_link(pm, n, i)
+        end
+    end
+
+end
+
+function add_bus_voltage_setpoint(sol, pm::QCWRTriPowerModel)
+    add_setpoint(sol, pm, "bus", "vm", :vm)
```
ccoffrin: A better fix to this issue would be to generalize the current `add_bus_voltage_setpoint(sol, pm::QCWRPowerModel)` to a function like `add_bus_voltage_setpoint{T <: QCWRForm}(sol, pm::GenericPowerModel{T})`

test/opf.jl
```diff
@@ -217,6 +217,35 @@ end
...
+
+        @test result["status"] == :LocalOptimal
+        @test isapprox(result["objective"], 11567.1; atol = 1e0)
+        @test isapprox(result["solution"]["bus"]["1"]["va"], 0.0; atol = 1e-4)
+        @test isapprox(result["solution"]["bus"]["4"]["va"], 0.0; atol = 1e-4)
+    end
+    @testset "24-bus rts case" begin
+        result = run_opf("../test/data/case24.m", QCWRTriPowerModel, ipopt_solver)
+
+        @test result["status"] == :UserLimit
```
ccoffrin: Seems strange this would be hitting the user limit.  Maybe an indication of a bug.  If not a bug, in the interest of the test runtime, it is better to omit this test.

src/form/wr.jl
```diff
@@ -736,3 +737,97 @@ function constraint_power_magnitude_link_on_off(pm::QCWRPowerModel, n::Int, i, a
...
+    v = pm.var[:nw][n][:vm]
+    t = pm.var[:nw][n][:va]
+
+    td = pm.var[:nw][n][:td]
+    si = pm.var[:nw][n][:si]
+    cs = pm.var[:nw][n][:cs]
+
+    w = pm.var[:nw][n][:w]
+    wr = pm.var[:nw][n][:wr]
+    lambdar = pm.var[:nw][n][:lambdar]
```
ccoffrin: might consider calling these `lambda_wr` and `lambda_wi`, to connect them more explicitly to the `wr` and `wi ` variables

test/opf.jl
```diff
@@ -235,8 +235,8 @@ end
 
         @test result["status"] == :LocalOptimal
         @test isapprox(result["objective"], 11567.1; atol = 1e0)
-        @test isapprox(result["solution"]["bus"]["1"]["va"], 0.0; atol = 1e-4)
-        @test isapprox(result["solution"]["bus"]["4"]["va"], 0.0; atol = 1e-4)
+        @test isapprox(result["solution"]["bus"]["1"]["va"], 0.05285; atol = 1e-4)
+        @test isapprox(result["solution"]["bus"]["4"]["va"], 0.03252; atol = 1e-4)
     end
```
ccoffrin: This is highly suspicious.   Bus 1 and 4 are references buses.  va should be fixed to 0 by https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/prob/opf.jl#L30.

kaarthiksundar: Let me try to figure out where the bug is. The thing is it is not clear to me as to where to start debugging. The convex hull formulation is correct; I have verified it multiple times. My suspicion is with the other parts of the qc code. Do you have a suggestion as to where to start with?

kaarthiksundar: Okay, I figured it out. All the QC constraints need to be templated. I shall work on getting that done.

ccoffrin: Good catch.  You should revisit the 24-bus test once that is done, probably it will work :-)

ccoffrin:
Looks good to me.  @kaarthiksundar you ready for a merge?

kaarthiksundar:
Yes @ccoffrin

PR #195, opened 2017-11-15
Bug fix on QC-OTS formulation

kaarthiksundar:
````relaxation_product_on_off(pm.model, vm_fr[i], vm_to[j], vv[l], z[l])```` becomes  ````relaxation_product_on_off(pm.model, vm_fr[l], vm_to[l], vv[l], z[l])````


ccoffrin:
Based on CI, it looks like post-fix some objective values in the tests need to be updated.

also should add a note to the changelog.

PR #196, opened 2017-11-15
QC-OTS tests and changelog

kaarthiksundar:


ccoffrin:
Linux tests pass, so merging.

PR #198, opened 2017-11-16
QC exact reformulation for square term

kaarthiksundar:
Added exact reformulation for ````y = x*x```` using a lambda formulation.


ccoffrin:
Looks suspicious to me.  @hhijazi what do you think?

ccoffrin:
Also could you merge master into your fork.  There are some repeat commits in here, which make it hard to see what is being changed.

kaarthiksundar:
My bad. Agreed.

-- 
Kaarthik Sundar

hhijazi:
@ccoffrin, seems ok to me, this is a lifted convex hull formulation. @kaarthiksundar I was not able to see the lifted variable representing x*y in the commit log, maybe it's already in the code?

ccoffrin:
This morning we figured out that the quadratic lower bound is missing.  Compare to this, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/relaxation_scheme.jl#L70

PR #201, opened 2017-12-09
adding juniper for ac minlp tests

ccoffrin:


ccoffrin:
@Wikunia I noticed that on this 5-bus example Juniper has some issues.  To me it looks like bad performance of ipopt using mumps, but maybe you can suggest some settings to make it more robust.

Wikunia:
Hmm it looks like Ipopt isn't able to find any feasible solution even after several restarts. Another question. Is the problem in `tnep` a minlp  or only a nlp? Seems like juniper "thinks" it's only a nlp.

ccoffrin:
Yes I noticed that as well, but if I turn down the optimality tolerance of ipopt, it seems to do better?

`tnep` should have discrete switching variables on the lines, however, it may be integral at the root node, which is fine, but a harder test case would be ideal.  At this point I am just trying to reproduce the tests as they were when using Bonmin.

Wikunia:
I think I have to dive deeper into PowerModels. Atm I am a bit at a loss. It seems like juniper is started more than once (is that correct?) sometimes having no integer variables (are you solving the relaxation first?). And the Juniper model object seems to get corrupted somewhere as `nl_solver` should be ipopt but the logs show me that it is Juniper itself.

Wikunia:
Yes it's weird that it works without setting the tolerance which defaults to `1e-8` but doesn't find a feasible solution with a worse tolerance of `1e-4`. Btw I am using ma27 not mumps for the test.

ccoffrin:
If we have high confidence that the reason for the issue is the NLP solver and we don't have any clever ideas for work arounds I will change the case to a more friendly one (e.g. case14).  However in the past I never had any issue solving this case with bonmin, so maybe its an example where bonmin is doing something clever / worthwhile...

Wikunia:
Is there a reason why the tolerance is reduced to `1e-4` ? For me it looks like it that stage it has nothing to do with Juniper. Is it possible to get the JuMP model of that problem easily? Then maybe I can run the relaxation outside of Juniper just with Ipopt to be sure that it is an Ipopt problem?

ccoffrin:
Here is how you can get the JuMP model,
```
pm = build_generic_model("case5.m", ACPPowerModel, PowerModels.post_ots)
m = pm.model
```

Wikunia:
This might need a second look. I changed 
https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/variable.jl#L337 to `:Cont`

then used 
```
pm = build_generic_model("case5.m", ACPPowerModel, PowerModels.post_ots)
m = pm.model
setsolver(m, IpoptSolver(tol=1e-4))
solve(m)
```
This returns `:Optimal`

Wikunia:
Ah okay the problem appears only on `case5_tnep.m` on my machine (with ma27) and there Ipopt returns infeasible.

ccoffrin:
For future reference you can use `solve(m, relax=true)` to ask JuMP to solve a continuous relaxation of a discrete problem.

Wikunia:
Haha true I forgot that option.

Issue #210, opened 2018-03-01
Cleaner Matpower Parser


ccoffrin:
Break `parse_matpower` into two distinct steps.

1. build a 1-to-1 dict of the matpower datafile
2. translate the matpower dict into a PowerModels dict

PR #215, opened 2018-03-05
Refactored Matlab/Matpower Parser

ccoffrin:
Address #210, prepares for shared Matlab data parsing.

### DIFF
--- a/src/io/common.jl
+++ b/src/io/common.jl
@@ -1,9 +1,9 @@
 ""
 function parse_file(file::String)
     if endswith(file, ".m")
-        pm_data = PowerModels.parse_matpower(file)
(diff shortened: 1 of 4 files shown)


src/io/common.jl
```diff
@@ -1,9 +1,9 @@
 ""
 function parse_file(file::String)
     if endswith(file, ".m")
-        pm_data = PowerModels.parse_matpower(file)
+        pm_data = parse_matpower(file)
     else
-        pm_data = PowerModels.parse_json(file)
+        pm_data = parse_json(file)
```
yeesian: do we have any tests that covers this portion of the parser?

PR #219, opened 2018-03-13
Memento Logging

### LINKED ISSUE #218: Memento Logging?
Any interest in switching over to using [Memento.jl](https://github.com/invenia/Memento.jl) for logging? We're finding that PowerModels.jl throws a lot of warning under some specific circumstances and it would be nice if we could configure PowerModels.jl logging on the application side. We're just using [Suppressor.jl](https://github.com/Ismael-VC/Suppressor.jl) for now, but it would be nice if we could have a more flexible solution.

rofinn:
Updates to log message to a package level Memento logger. Closes #218.

- Required updating the minimum julia version as Memento dropped 0.5 support a while ago. While we could use an older version of Memento on 0.5 we'd get a lot of deprecation message on 0.6.
- Updated the minimum Compat version to support `@__MODULE__` on 0.6/0.7.
- We define a global `LOGGER` constant for the package and pass it as the first argument to all the existing logging calls.
- The package level logger can be set to `"error"` (via `setlevel!(getlogger(PowerModels), "error")`) to suppress warning messages without conflicting with warning messages from other libraries.

NOTES:

- Might want to convert a lot of the print statements in the code to `debug` calls.
- Seems like some of the regular warning messages could be `notice` messages.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6
 JSON 0.6.0
 MathProgBase 0.5.4
 JuMP 0.17 0.19-
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,4 +2,5 @@ julia 0.6
 JSON 0.6.0
 MathProgBase 0.5.4
 JuMP 0.17 0.19-
-Compat 0.17
+Compat 0.26.0
+Memento 0.5.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 matrix:
   allow_failures:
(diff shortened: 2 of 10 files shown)


REQUIRE
```diff
@@ -1,5 +1,6 @@
-julia 0.5
+julia 0.6
 JSON 0.6.0
 MathProgBase 0.5.4
 JuMP 0.17 0.19-
-Compat 0.17
+Compat 0.26.0
```
ccoffrin: Did you check this is the min version of compat?  I often get pinged on this when tagging new versions.

rofinn: Yeah, [0.26.0](https://github.com/JuliaLang/Compat.jl/releases/tag/v0.26.0) was when `@__MODULE__` was added to Compat which is the only reason I needed to bump the version number.

PR #220, opened 2018-03-13
FIX: Use of deprecated syntax

pseudocubic:
Requires complete drop of support for Julia v0.5, but prepares for eventual
Julia v0.7 update. Because v0.5 tests are currently allowed failures in travis, 
and will be removed if PR #219  is accepted, it has not been removed in this PR.

Addresses linter warning "Use of deprecated parameter syntax" by changing 
e.g. `function some_name{T}(var)` to `function some_name(var) where T`

Also addresses linter warning "Use of deprecated `type` syntax" which was 
deprecated in favor of `mutable struct`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 matrix:
   allow_failures:
     - julia: nightly
-    - julia: 0.5
 cache:
  directories:
    - /home/travis/.julia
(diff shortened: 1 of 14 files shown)


ccoffrin:
Looks good to me.  We will drop support for v0.5.  You also need to check if Compat is still necessary. If so, determine the minimum version, and increase if it is larger than that of #219.

pseudocubic:
@ccoffrin Since `Compat` doesn't seem necessary at all for anything now that Julia v0.5 support is being dropped, I decided to completely remove it in this PR, and #219 can add it back separately for their use of `@__MODULE__`

Issue #221, opened 2018-03-15
Tranformer Direction Bug


ccoffrin:
Need to fix a bug in, [check_branch_directions](https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/data.jl#L439), when there are two parallel transformers are given in different directions the admittance values need to be updated as well.

PR #226, opened 2018-03-19
ADD: Added support for parsing of PTI files

### LINKED ISSUE #211: Parser for PSSE Files
Turn PSSE text files into a Julia dictionary.

pseudocubic:
Added support for the parsing of PTI v33+ raw files via function
`parse_pti`, as defined by the PSS(R)E 33.3 Program Operation Manual.

Parser supports all defined sections *except* "GNE Device Data". Data in
each section is parsed into its correct data type, i.e. `Int64`,
`Float64`, or `String`.

The output data is a `Dict` from `parse_pti` and has the following
format:

    Dict{String, Array{Dict{String, Any}}}

Each element/line of the PTI is stored as an element of the `Array` in
their appropriate Section given by the outer `String`. Each element/line
is separated into subelements parsed into their appropriate data types
into the inner `Dict` with field names given by the innermost `String`.

Unit tests check to ensure that all elements from known test files are
parsed into a `Dict` with the above format, and can be found in
`test/pti.jl`.

Parsing of the PTI data to a PowerModels compatible format is *not*
yet supported by this update.

`CHANGELOG.md` and documentation in `docs/src/parser.md` have been
updated to reflect these changes.

Version of Julia in `REQUIRE` has been bumped to `0.6` to reflect other
recent changes already merged into `master`, i.e. #219, #220 and #221.

Closes #211


ccoffrin:
Looks great!  

After a quick pass my only recommendation would be to give the very short models, e.g. `frankenstein_00c.raw`, a more semantic name, like `parser_test_a.raw`.  The frankenstein cases are designed to be component level tests that are mathematically solvable.

Issue #26, opened 2016-09-11
Standardize Variable Names


ccoffrin:
I propose the following a standard naming convention for the polar and rectangular variable spaces, let "v" be a complex number, then:
 - vm - magnitude of v
 - va - angle of v
 - vr - real component of v
 - vi - imaginary component of v

ccoffrin:
Another challenge is how to deal with variable products.  It has been convenient to use w for the product of two voltages (i.e. v).  But we need a more general rule for the product of other complex numbers.

I suggest if x and y are complex numbers then xy can be represented as,

- xym - suggests xm*ym
- xya - suggests xa + ya
- xyr - suggests xr + yr
- xyi - suggests xi + yi
 
Due to the convenient existence of w, which looks like vv, we can use it as a shorthand of these cases,

- wm - short for vvm
- wa - short for vva
- wr - short for vvr
- wi - short for vvi

ccoffrin:
In light of this convention, and that S is the standard value for complex power,  should p and q be renamed sr and si?

kaarthiksundar:
We can do the following: at instances where there is no standard agreed upon naming convention like p and q for sr and si, we can follow your proposed scheme.

rb004f:
I don't think we can rename p and q as sr and si.  To many people "know" what p and q mean, and to replace that may cause too much confusion.  I tend to agree with Kaarthik here, where we use sr and si whenever there is not an agreed upon notation.

ccoffrin:
If we adopt @kaarthiksundar's suggestion, what is the list of standard agreed upon names?

Another option is for some parameters like p and q to have two aliases for the same set of variables, the typical name and the standardized convention.  It's unclear to me if JuMP could handle that.

kaarthiksundar:
@ccoffrin I am pretty (95%) sure JuMP cannot handle that. I have been experimenting with variable names for quite some time and in my experience, that is not possible. 

I guess the only agreed upon names are p, q and w.

mlubin:
It could handle it with some hacks, but would need to be convinced that it's worth making the scoping more complicated and confusing than it already is. You could just use your own dictionary.

rb004f:
I think if we carefully document that p and q are the exception to the convention (and why), this will be ok.  w is a somewhat newer convention, so maybe that can be changed without too much heart burn?

ccoffrin:
I like the w notation and It's very nice in the SDP models. I think we have a plan.

[...]

PR #229, opened 2018-03-25
Adding SOC extended DistFlow (BFM) formulation

frederikgeth:
I'm working on an implementation of the extended DistFlow formulation.

Before I continue, I'm wondering what you think of the following issues:
- Adding a line index `i` to  `constraint_voltage_angle_difference(pm, n, *i*, f_bus, t_bus, buspair["angmin"], buspair["angmax"])`
- Name of the formulation, as it uses `W` variables but not WR/WI but squared current magnitude instead. Currently it is named `AbstractWIForm`, to refer to the 'W' voltage variables and current variables 'I', but I don't think it is an ideal name. 
- Place in the type hierarchy: `AbstractWForms = Union{AbstractWRForms, AbstractWIForm}`
- As it is a BFM, one equation defines both the to and from side active/reactive power flow. So the call to `constraint_ohms_yt_to(pm, i)` in this formulation does nothing. Should the problem definitions be changed, where a call `constraint_ohms(pm, i)` for BIM-type models itself calls `constraint_ohms_yt_from(pm, i)` and `constraint_ohms_yt_to(pm, i)`?
- The line power equations are prepared with support for asymmetric line shunts and shunt conductance.
- There's still something wrong with `constraint_voltage_angle_difference` though, as the test case 24 (with SAD) returns an objective value which is too low, I'll take a deeper look hopefully sometime the coming week. 

any other comments are welcome of course :)


src/core/variable.jl
```diff
@@ -342,11 +344,35 @@ end
...
+        lowerbound = 0,
         upperbound = 1,
         category = :Int,
         start = getstart(branches, l, "branch_tnep_start", 1.0)
     )
 end
+
+
+"variable: `0 <= i[l] <= (Imax)^2` for `l` in `branch`es"
+function variable_series_current_magnitude_sqr(pm::GenericPowerModel, n::Int=pm.cnw; bounded = true)
```
ccoffrin: I don't think you need this constraint.  You can move this implementation into this shared variable space,

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/wr.jl#L399

src/form/wi.jl
```diff
@@ -0,0 +1,139 @@
...
+
+"do nothing, this model does not have complex voltage constraints"
+function constraint_voltage(pm::GenericPowerModel{T}, n::Int) where T <: AbstractWIForm
+end
+
+
+"""
+Creates Ohms constraints (yt post fix indicates that Y and T values are in rectangular form)
+"""
+function constraint_ohms_yt_from(pm::GenericPowerModel{T}, n::Int, f_bus, t_bus, f_idx, t_idx, g, b, c, tr, ti, tm) where T <: AbstractWIForm
```
ccoffrin: I agree this does not make a lot of sense.  I think this formulation needs a new OPF formulation that better reflect its construction.

src/form/wi.jl
```diff
@@ -0,0 +1,139 @@
...
+    p_fr = pm.var[:nw][n][:p][f_idx]
+    q_fr = pm.var[:nw][n][:q][f_idx]
+    p_to = pm.var[:nw][n][:p][t_idx]
+    q_to = pm.var[:nw][n][:q][t_idx]
+    w_fr = pm.var[:nw][n][:w][f_bus]
+    w_to = pm.var[:nw][n][:w][t_bus]
+    l = f_idx[1]
+    i =    pm.var[:nw][n][:i][l]
```
ccoffrin: Have a look at the QC model for variable naming conventions.  I have been using `cm` so far.

ccoffrin:
Based on your recommendation lets adopt the name BranchFlow (i.e. `bf`) for this variant of the model.  

Because the equations in the base formulation are quite different, I think we need to add a new problem name for this variant of the model, I am thinking `opf_bf`,  `pf_bf` and so on.  The primary differences will be the addition of branch current variables and different constraints that are stated on the branches; these should be called something like `constraint_power_losses` and `constraint_voltage_magnitude_difference`.  In this case, we can re-use the existing formulation names (for example SOCWR and ACP) for different variants of this new problem, which will make re-using the shared functions simple. 

Concerning `constraint_voltage_angle_difference` and `constraint_voltage_magnitude_difference`, for consistency we should make both of them branch-wise constraints, similar to https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/constraint_template.jl#L316.  This means passing in `arc_from` in addition to `f_bus` and `t_bus`.

Concerning, the inconsistency on test case 24.  My guess is that it is due to the fact that the BF formulation does not implement this constraint, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/wr.jl#L516  To the best of my knowledge, the constraint has never been projected into the BF space.  I would be happy to do the derivation, in due course.  For now I would confirm that this is the issue and then integrate the case 24 test with a different objective value.

src/form/wi.jl
frederikgeth: ok, cm it is.

frederikgeth:
Thank you very much for your feedback. Further commits forthcoming.

I'll make the `opf_bf`, `pf_bf`problem types. The suggested name `constraint_power_losses` makes a lot of sense. However, I'm not sure the formulation itself should be called `bf` as this is just an example of a BFM formulation. I'm partial to DistFlow (the nonconvex quadratic one in p, q, vm), convexified DistFlow (the SOC one in p, q, w, (cm)^2), linearized DistFlow (LP in p, q, w). What do you think? How would you abbreviate?

Concerning `constraint_voltage_angle_difference` and `constraint_voltage_magnitude_difference`, I think the passing `arc_from` is a good solution. 

I will confirm soon that the difference between the current SOC-BIM and SOC-BFM is due to the missing lifted nonlinear cuts. For now I changed the objective value in the SOC-BFM tests.

frederikgeth:
Updated the naming of the current variable, now also documented [here](https://github.com/frederikgeth/PowerModels.jl/blob/cf15a85f9c5d00fa51f61237c40f818ff37ed9f7/docs/src/developer.md) for future reference. Is this what you expect cfr https://github.com/lanl-ansi/PowerModels.jl/issues/26?

`opf_pf` is now defined, including some code for compatibility of formulations between `opf`and  `opf_bf`in `shared`. `pf_bf`is still a todo.

The results of SOCWR and DistFlow for `opf` and `opf_bf` are not yet consistent for case 24, even after playing with inclusion/exclusion of `constraint_voltage_angle_difference` and `cut_complex_product_and_angle_difference`, I'll need to dig deeper to figure this out, sometime later this week.

ccoffrin:
Concerning the problem name.  We need a generic problem name which will then take different forms.  As an example, the abstract problem `opf_xyz` will become DistFlow with `APowerModel`, ConvexDistFlow with `BPowerModel`, and LinDistFlow with `CPowerModel`.

I am inclined to call the abstract problem as `opf_bf` (for branch flow) because it sounds more formulation agnostic to me than DistFlow, but I am open to other proposals.

The variable name docs looks good to me.  I like the `cc` notation for current magnitude squared.  It is more consistent with standard.

frederikgeth:
Ok, great, let me confirm. So, the problem types are named `opf_bf`and `pf_bf`, with corresponding filenames and function names. 

Currently, the convex relaxation of the (balanced) DistFlow formulation (shorthand `DF`) is type `SOCDFPowerModel`, contained in `df.jl`. If we eventually implement the nonconvex formulation it could be called `DFPowerModel` and the linearized version could be `LinDFPowerModel`. 

The BFM-equivalent constraints of BIM-style `constraint_ohms_yt_from` and `constraint_ohms_yt_to` are currently conceived as:
- constraint_branch_flow_losses: defining the relationship between `p_fr`, `p_to` and the total line loss (and equivalent for `q`)
- constraint_branch_kvl: defining the relationship between the complex voltages at the from and the to side, whether in normal (complex) variables, or squared variables
- constraint_branch_current: defining how the branch current relates to the (complex) power flow and the voltage magnitudes

src/form/wi.jl
frederikgeth: ok, I agree, made a proposal that still allows you to use BIM formulations in the BFM problem type and vice-versa.

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+    for i in ids(pm, :bus)
+        constraint_kcl_shunt(pm, i)
+    end
+
+    for i in ids(pm, :branch)
+        constraint_branch_flow_losses(pm, i)
+        constraint_branch_kvl(pm, i)
+        constraint_branch_current(pm, i)
+
+        #constraint_voltage_angle_difference(pm, i)
```
ccoffrin: Why is this commented out?

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+""
+function run_opf_bf(file, model_constructor, solver; kwargs...)
+    return run_generic_model(file, model_constructor, solver, post_opf_bf; kwargs...)
+end
+
+""
+function post_opf_bf(pm::GenericPowerModel)
+    variable_voltage(pm)
+    variable_generation(pm)
+    variable_branch_flow(pm)
```
ccoffrin: We should add branch current variables as well, right?

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+    for i in ids(pm, :ref_buses)
+        constraint_theta_ref(pm, i)
+    end
+
+    for i in ids(pm, :bus)
+        constraint_kcl_shunt(pm, i)
+    end
+
+    for i in ids(pm, :branch)
+        constraint_branch_flow_losses(pm, i)
```
ccoffrin: `constraint_branch` feels redundant given that losses are only defined on branches.  How about `constraint_flow_losses`?

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+        constraint_branch_kvl(pm, i)
```
ccoffrin: kvl is a property of cycles on the network, correct?  I don't see how it connects to this constraint.  I think `voltage_magnitude_difference` is accurate and consistent with the constraints already defined.

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+        constraint_branch_current(pm, i)
```
ccoffrin: This is probably the exception to the rule.  Here I think it makes sense to keep `branch` explicit, because there are so many possibilities for different currents.

Also I will have to think about how to unify this with `constraint_voltage`, but I will leave that for future issue that is outside the scope of this PR.

src/prob/opf_bf.jl
```diff
@@ -0,0 +1,51 @@
...
+""
+function post_opf_bf(pm::GenericPowerModel)
+    variable_voltage(pm)
+    variable_generation(pm)
+    variable_branch_flow(pm)
+    variable_dcline_flow(pm)
+
+    objective_min_fuel_cost(pm)
+
+    constraint_voltage(pm)
```
ccoffrin: It is not immediately clear to me if this is needed in problem definition.  It might help you to comment it out for now.

src/form/shared.jl
```diff
@@ -3,22 +3,52 @@
...
-# Note that Union types are discouraged in Julia, 
+# Note that Union types are discouraged in Julia,
 # https://docs.julialang.org/en/release-0.6/manual/style-guide/#Avoid-strange-type-Unions-1
 # and should be used with discretion.
 #
 # If you are about to add a union type,
 # first double check if a different type hierarchy can resolve the issue
 # instead.
 #
+AbstractBIMForms = Union{AbstractACPForm, AbstractACTForm, AbstractDCPForm, AbstractWRForm, AbstractWRMForm}
```
ccoffrin: Note that type unions are discouraged in Julia.  So for now, I would recommend removing this collection and simply say that the `opf_bf` problem does not yet support these forms and defining new forms just for BFMs.

Later on after this PR, I will consider unification of both forms.

src/core/constraint_template.jl
```diff
@@ -439,11 +439,12 @@ function constraint_voltage_angle_difference(pm::GenericPowerModel, n::Int, i::I
     branch = ref(pm, n, :branch, i)
     f_bus = branch["f_bus"]
     t_bus = branch["t_bus"]
+    arc_from = (i, f_bus, t_bus)
     pair = (f_bus, t_bus)
     buspair = ref(pm, n, :buspairs, pair)
 
     if buspair["branch"] == i
```
ccoffrin: This might also be the issue of inconsistent on case 24.  You might be to post this constraint for all branches in the BFM models.

src/prob/opf_bf.jl
frederikgeth: current is also a flow variable, so I had put the initialization of the current variables at the same level as the calls to the active and reactive power initialization. will be addressed

src/prob/opf_bf.jl
frederikgeth: forgot to remove after trying to find the source of the issue with case24 optimality BIM vs BFM

src/prob/opf_bf.jl
frederikgeth: ok

src/prob/opf_bf.jl
frederikgeth: Maybe it should refer to Ohm's law, as it relates two voltages through impedance. But that may create confusion w.r.t. BIM-style OPF problem formulation.

src/prob/opf_bf.jl
frederikgeth: ok

src/form/shared.jl
frederikgeth: ok, unification code removed

src/prob/opf_bf.jl
frederikgeth: ok, removed

frederikgeth:
Results are now identical between `SOCWRForm` and `SOCDFForm`, for both opf and pf problem types, including for case24. Unit tests are added for `opf_bf` and `pf_bf` for `SOCWRForm` and problem definition documentation is updated.

src/core/constraint_template.jl
frederikgeth: In the end, this was not the problem, just a typo in the DistFlow equation for `constraint_voltage_angle_difference`

src/form/df.jl
```diff
@@ -0,0 +1,163 @@
...
+"default SOC constructor"
+SOCDFPowerModel(data::Dict{String,Any}; kwargs...) = GenericPowerModel(data, SOCDFForm; kwargs...)
+
+""
+function variable_branch_current(pm::GenericPowerModel{T}, n::Int=pm.cnw; kwargs...) where T <: AbstractDFForm
+    variable_branch_series_current_magnitude_sqr(pm, n; kwargs...)
+end
+
+""
+function variable_branch_flow(pm::GenericPowerModel{T}, n::Int=pm.cnw; kwargs...) where T <: AbstractDFForm
```
ccoffrin: I think this function can be omitted.  The most generic version works for all GenericPowerModels.

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/variable.jl#L223

src/form/df.jl
```diff
@@ -0,0 +1,163 @@
...
+    variable_reactive_branch_flow(pm, n; kwargs...)
+end
+
+""
+function variable_voltage(pm::GenericPowerModel{T}, n::Int=pm.cnw; kwargs...) where T <: AbstractDFForm
+    variable_voltage_magnitude_sqr(pm, n; kwargs...)
+end
+
+"do nothing, this model does not have complex voltage constraints"
+function constraint_voltage(pm::GenericPowerModel{T}, n::Int) where T <: AbstractDFForm
```
ccoffrin: This implementation can be provided without any constraint on `T` and can be placed in `core/constraint.jl`

src/form/df.jl
```diff
@@ -0,0 +1,163 @@
...
+function constraint_voltage_magnitude_difference(pm::GenericPowerModel{T}, n::Int, i, f_bus, t_bus, f_idx, t_idx, r, x, g_sh_fr, b_sh_fr, tm) where T <: AbstractDFForm
+    p_fr = pm.var[:nw][n][:p][f_idx]
+    q_fr = pm.var[:nw][n][:q][f_idx]
+    w_fr = pm.var[:nw][n][:w][f_bus]
+    w_to = pm.var[:nw][n][:w][t_bus]
+    ccm =  pm.var[:nw][n][:ccm][i]
+
+    #define series flow expressions to simplify Ohm's law
+    p_fr_s = p_fr - g_sh_fr*(w_fr/tm^2)
+    q_fr_s = q_fr + b_sh_fr*(w_fr/tm^2)
```
ccoffrin: It's not immediately obvious to me that this is correct.  Did you double check?  Until we have an asymmetrical network test this will not be checked by unit tests.

src/form/df.jl
```diff
@@ -0,0 +1,163 @@
...
+"""
+function constraint_branch_current(pm::GenericPowerModel{T}, n::Int, i, f_bus, f_idx, g_sh_fr, b_sh_fr, tm) where T <: AbstractDFForm
+    p_fr = pm.var[:nw][n][:p][f_idx]
+    q_fr = pm.var[:nw][n][:q][f_idx]
+    w_fr = pm.var[:nw][n][:w][f_bus]
+    ccm  = pm.var[:nw][n][:ccm][i]
+
+    #define series flow expressions to simplify constraint
+    p_fr_s = p_fr - g_sh_fr*(w_fr/tm^2)
+    q_fr_s = q_fr + b_sh_fr*(w_fr/tm^2)
```
ccoffrin: Same as above, it's not immediately obvious to me that this is correct. Did you double check?  Especially because in my derivation the shunt value did not occur in this constraint.

src/form/df.jl
```diff
@@ -0,0 +1,163 @@
...
+    @constraint(pm.model,
+        tan(angmax)*(( tr - tzi*b_sh_fr)*(w_fr/tm^2) - tzr*p_fr - tzi*q_fr)
+                 >= ((-ti - tzr*b_sh_fr)*(w_fr/tm^2) - tzr*q_fr + tzi*p_fr)
+        )
+
+end
+
+
+"variable: `0 <= i[l] <= (Imax)^2` for `l` in `branch`es"
+function variable_branch_series_current_magnitude_sqr(pm::GenericPowerModel, n::Int=pm.cnw; bounded = true)
```
ccoffrin: Can we replace this function with the one that is already defined here, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/wr.jl#L399?  If so, that function can move to core/variable.jl because it is generally applicable.

src/form/df.jl
frederikgeth: indeed

src/form/df.jl
frederikgeth: ok

src/form/df.jl
frederikgeth: The derivation goes as follows. Branches are modeled with an ideal transformer and a pi-section in series.

As the ideal transformer is located at the from side, the voltage after the transformer, i.e. at the from side of the pi-element following it, is actually `(w_fr/tm^2)`. Furthermore, as this transformer is ideal, it is lossless. Moreover, as we'll be doing the DistFlow derivation, which performs the angle relaxation, transformer phase shift does not influence the power flow (except for `constraint_voltage_angle_difference`.)

We observe that a a branch's power flow balance with asymmetrical shunts is composed of three loss contributions: `p_fr + p_to = p_sh_loss_fr + p_s_loss + p_sh_loss_to`, `q_fr + q_to = q_sh_loss_fr + q_s_loss + q_sh_loss_to`, where `x_sh_loss_fr`, `x_sh_loss_to` are the to from and to side shunt losses, and `x_s_loss` is the series flow loss. We can further define the flow only through the series impedance as `p_fr_s + p_to_s = p_s_loss`, `q_fr_s + q_to_s = q_s_loss`. 

The pi-section loss terms, expressed in the `w` and `ccm` variables therefore are:
- `p_sh_loss_fr =  g_sh_fr*(w_fr/tm^2)`
- `q_sh_loss_fr = -b_sh_fr*(w_fr/tm^2)`
- `p_s = r*ccm` 
- `p_s = x*ccm`
- `p_sh_loss_to =  g_sh_to*(w_to)`
- `q_sh_loss_to = -b_sh_to*(w_to)`

Ohm's law only relates the voltage magnitude and the series current and series impedance of the pi-element. Therefore, after performing the DistFlow derivation (`v_j = v_i - z_ij *I_ij_s`, with RHS and LHS multiplied by their complex conjugated) and the subsititution of the squared node voltage magnitude, (with `w` and squared branch *series* current magnitude, with `ccm`), you obtain expressions in the *series* complex power flow variables `p_fr_s, q_fr_s`. Furthermore, in these variables `p_fr_s^2 +q_fr_s^2 == (w_fr/tm^2)*ccm` is a valid expression, which is then relaxed. 

I can add this somewhere in the documentation.

I haven't found the time to rederive `constraint_voltage_angle_difference` to confirm its current implementation though.

src/form/df.jl
frederikgeth: These constraints have different meaning: bounding series current in this BFM derivation vs bounding total current magnitude in QC. Anyway, I updated the series current upper bound with a compensation term for the current contribution of the shunts.

docs/src/developer.md
```diff
@@ -1,3 +1,113 @@
...
+### Impedance
+
+Defining impedance
+$z = r + j\cdot x$:
+- `r`: resistance ($\Omega$)
+- `x`: impedance ($\Omega$)
+
+## DistFlow derivation
+
+### For an asymmetric pi section
```
frederikgeth: here's the derivation, @ccoffrin

ccoffrin:
The implementation looks reasonable.   I would like to understand the math a bit better before we merge this in.  Can we setup a VC to go over it?

ccoffrin:
@frederikgeth, fixed merge conflict so that travis would run.

ccoffrin:
@frederikgeth seems there are still a few updates needed to bring thus up to date with master.

ccoffrin:
Thanks!

PR #235, opened 2018-04-02
Network Topology Processing

### LINKED ISSUE #207: Functions for Network Processing
Add functions for updating the network topology status and also selecting connected components.

ccoffrin:
Adding some functions for network topology processing.  

@rofinn, I set the default logging level to "info" as this is more consistent with Julia's default behavior and the info data should be present for novice users.

Closes #207


rofinn:
I recommend leaving logging configuration up to the application using your package, but since I haven't gotten around to solving invenia/Memento.jl#63 this seems like a reasonable compromise.

PR #238, opened 2018-04-02
FIX: Load/Shunt "status" value Type

### LINKED ISSUE #234: Status Value Consistency
Load and shunt status values are currently boolean `true/false` the standard in most other components and formats is `1/0`.  These should be made consistent, I think `1/0` will be more intuitive for the majority of users.

pseudocubic:
Changes the "status" value of Loads and Shunts to `1` or `0` instead of
`true` or `false`, by using `convert(Int8, value)` during the
construction in `split_loads_shunts()`.

Refactors relevant lines to make lines more readable.

Closes #234

### DIFF
--- a/src/core/data.jl
+++ b/src/core/data.jl
@@ -670,13 +670,13 @@ function _propagate_topology_status(data::Dict{String,Any})
     incident_load = bus_load_lookup(data["load"], data["bus"])
     incident_active_load = Dict()
     for (i, load_list) in incident_load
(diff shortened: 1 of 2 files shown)


ccoffrin:
Some functions in data.jl will need to be updated so that types are consistent.

For example, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/data.jl#L851

Issue #213, opened 2018-03-01
Data Formats Documentation


ccoffrin:
Improve data format documentation to reflect different file formats and data models.  Provide some examples in the quick-start guide for raw parsing of matpower and psse data.

Issue #212, opened 2018-03-01
PSSE dict to PowerModels dict


ccoffrin:
Translation function from the PSSE dict to the PowerModels dict.

ccoffrin:
Regarding this feature, I recomend we implement it in two steps.  First focus on support for
- bus
- load
- shunt (fixed and switched)
- branch
- transformer (two-winding)
These features will cover the bulk of our test cases.

Once that is done and well tested, then then we can move on to,
- transformer (three-winding)
- HVDC lines

ccoffrin:
Because PTI ids are complex, I also recommend adding a `pti_id` field to each PowerModels component.  To be compatible with JSON `pti_id` can be an array with a mix of ints and strings, which depend on the source component.

pseudocubic:
bus, load, shunt, generator, branch, two-winding and three-winding transformers incl. magnetizing admittance are now supported. 

HVDC lines remaining

PR #241, opened 2018-04-05
ADD: Added parsing of PSS(R)E data into PowerModels data

pseudocubic:
Added support for the parsing of PTI v33+ raw files via function
`parse_psse`, as defined by the PSS(R)E 33.3 Program Operation Manual,
into a PowerModels format suitable for power network optimization.

The following network components are currently supported:
- bus
- load
- shunt
  - fixed
  - switched (approximated with default `gs=0.0`)
- generator
- branch
- transformer (two-winding)

Reasonable defaults for PowerModels-expected values which are not
supported by the PSS(R)E specification are assigned.

Because cost functions are not supported by the PSS(R)E specification,
`dcline` and `gen` components are given default cost function values,
namely:

    model = 2
    ncost = 3
    startup = 0.0
    shutdown = 0.0
    costs = [0.0, 1.0, 0.0]

Warnings are produced when attempting to use major unsupported
features, i.e. dclines transformers with magnetizing admittance, or
three-winding transformers. These features are still in development,
and will appear in future versions.

Added unit tests to compare PTI files translated via PowerWorld from
existing Matpower .m test files. There are several cases where
the `"objective"` is not expected to match, specifically, where the
Matpower files contain dclines (feature not currently supported by
the PTI parser), and where Matpower branches contain `angmin` and
`angmax` smaller than ~30deg, because this data field is not
supported by the PSS(R)E specification. Also, because cost functions
are not supported by the PSS(R)E specification, cost functions of the
imported Matpower data are set to match the defaults assigned to
imported PSS(R)E data before tests are run, to ensure consistency.

PowerModels version is added to the data structure, and `file_type`
and `file_version` are added to reflect the imported file type, i.e.
`matpower` or `pti`, and version of that file, respectively.

Added ability of matpower parser to parse `inf` values.

`CHANGELOG.md` and documentation in `docs/src/parser.md` have been
updated to reflect these changes.

Updated `CHANGELOG.md` to prepare for tagging of PowerModels v0.6.0
release.

Partially addresses #212 and #213


src/io/matpower.jl
```diff
@@ -147,11 +147,12 @@ function parse_matpower_string(data_string::String)
         case["name"] = "no_name_found"
     end
 
+    case["file_type"] = "matpower"
```
ccoffrin: what do you think about `source_type`, instead of file type?

src/io/matpower.jl
```diff
@@ -147,11 +147,12 @@ function parse_matpower_string(data_string::String)
...
     if haskey(matlab_data, "mpc.version")
-        case["version"] = matlab_data["mpc.version"]
+        case["file_version"] = VersionNumber(matlab_data["mpc.version"])
```
ccoffrin: `source_version`?

Issue #231, opened 2018-03-28
Add unit tests for PSSE features


pseudocubic:
- [x] Add unit tests for asymmetrical line charging added in #230 
- [x] Add unit tests for two-winding transformers with magnetizing admittance
- [x] Add unit tests for three-winding transformers
- [x] Add unit tests for three-winding transformers with magnetizing admittance
- [x] Add unit tests for two-terminal dc lines
- [x] Add unit tests for voltage source converter dc lines

Depends on #212

ccoffrin:
@pseudocubic looks like a good time to revisit this, everything but dc lines is implemented.

PR #254, opened 2018-04-19
ADD: Support PTI Three-Winding Transformers

pseudocubic:
Adds support for Three-winding transformers in PTI files. Transformers
are modeled by creating a "hidden star point bus", and creating a new
transformer branch for each winding of the transformer. For details
of mathematical formulation of the three transformer branches, see
"Power System Stability and Control" by Prabha Kundur, ISBN
0-07-035958-X, Sec. 6.2, Pg. 241, Eqn. 6.72.

Also adds ability to perform unit conversions in the cases that
values of the transformer WIND, R, and X, are not in the base units
defined by the buses and system SBASE. Applies to both two-winding
and three-winding transformers.

Adds unit tests for Three-winding Transformer. Created a Three-winding
transformer system in a PTI file, and used matpower and PowerWorld to
subsequently translate to matpower-type files for comparison.

Updated relevant documentation and CHANGELOG.

Refactors loops inside functions that parse components to eliminate
the enumeration variable (not needed).

Partially adddresses #212 and #231


test/psse.jl
```diff
@@ -173,4 +170,23 @@ end
         setlevel!(getlogger(PowerModels), "error")
     end
 
+    @testset "three-winding transformer" begin
+        data_pti = PowerModels.parse_file("../test/data/pti/three_winding_test.raw")
+        data_mp = PowerModels.parse_file("../test/data/matpower/three_winding_test_mp.m")  # Matpower Translation of PTI file
+        data_pw = PowerModels.parse_file("../test/data/matpower/three_winding_test_pw.m")  # PowerWorld Translation of PTI file
```
ccoffrin: What utility do you think the the Matpower translations of `three_winding_test.raw` give us?  I am inclined to omit these.

test/psse.jl
```diff
@@ -173,4 +170,23 @@ end
...
+
+        set_costs!(data_mp)
+        set_costs!(data_pw)
+
+        result_pti = PowerModels.run_opf(data_pti, PowerModels.ACPPowerModel, ipopt_solver)
+        result_mp = PowerModels.run_opf(data_mp, PowerModels.ACPPowerModel, ipopt_solver)
+        result_pw = PowerModels.run_opf(data_pw, PowerModels.ACPPowerModel, ipopt_solver)
+
+        @test result_pti["status"] == :LocalOptimal
```
ccoffrin: In addition to solving an OPF, I would also do some checks on the data it self.  For example, the number of buses and transformers in the PowerModels data model.

test/psse.jl
pseudocubic: I was doing internal testing before I got the powerworld translations and neglected to remove it.

ccoffrin:
As discussed on the phone lets,
- hardcode the expected outputs of .raw PF and OPF solves
- drop the .m files

pseudocubic:
@ccoffrin I've changed the tests, as we discussed

PR #230, opened 2018-03-27
UPD: Branch model support for asymmetrical line charging

pseudocubic:
Updated the Branch model to support asymmetrical line charging by
allowing for full complex variable in Ohms constraint equations, e.g.:
(from `form/acp.jl:90::91`)

    p[f_idx] ==  g/tm*v[f_bus]^2 +
        (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) +
        (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))
    q[f_idx] == -(b+c/2)/tm*v[f_bus]^2 -
        (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) +
        (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))

to

    p[f_idx] ==  (g+g_fr)/tm*v[f_bus]^2 +
        (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) +
        (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))
    q[f_idx] == -(b+b_fr)/tm*v[f_bus]^2 -
        (-b*tr-g*ti)/tm*(v[f_bus]*v[t_bus]*cos(t[f_bus]-t[t_bus])) +
        (-g*tr+b*ti)/tm*(v[f_bus]*v[t_bus]*sin(t[f_bus]-t[t_bus]))

where `c` is defined as `"br_b"`, or the total line charging susceptance
from the Matpower parser, and `g_fr` and `b_fr` are the real and
imaginary parts of the full complex admittance of the line shunt at the
"from"-bus end of the branch, respectively.

The Matpower parser now splits `"br_b"` into `b_fr = br_b / 2` and
`b_to = br_b / 2`, and assigns default values `g_fr = 0` and `g_to = 0`.

The constraints

- `constraint_loss_lb`
- `constraint_power_magnitude_link`
- `constraint_power_magnitude_link_on_off`

were changed to fail if `assert(g_fr == 0 && g_to == 0)`  fails, and
assigns  `c = b_fr + b_to`. These constraints will be re-derived from
first principles in the future.

Unit tests required no change.

Closes #214

PR #256, opened 2018-04-20
ADD: Support for transformer magnetizing admittance

### LINKED ISSUE #255: Dependency on DataStructures.jl
I just noticed this dependency on DataStructures.jl
https://github.com/lanl-ansi/PowerModels.jl/blob/e02b5143b4a2409312aedd787bfe90c7c63b4a33/src/io/pti.jl#L6

It is a best practice to put `using` calls in the root of the package 
https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/PowerModels.jl

Also the base `REQUIRE` should include all package dependencies.

In this case, I am wondering if it is worthwhile to add the dependency on DataStructures rather than the data structures that are in Base.

pseudocubic:
Adds support for non-zero magnetizing admittance parameters in PTI files
given via the MAG1 and MAG2 parameters in the transformer entries.
Feature is implemented by a front-side complex admittance, support for
which was added in #230, by setting `g_fr = MAG1` and `b_fr = MAG2`
on its associated transformer branch. For three-winding transformers,
the complex admittance is only added to the first of the three branches
that are created for each of the three windings.

Unit tests are added to check for consistency with PowerWorld PF
solutions, and to ensure OPF feasibility.

Documentation and changelog have been updated.

Added DataStructures to REQUIRE, and import within PowerModels.jl.

Addresses #212 and #231

Closes #255


ccoffrin:
Feature add looks good.  #255 is still under discussion.

Issue #242, opened 2018-04-06
PowerModels Data Extensions from PTI Data Files 


ccoffrin:
Figure out how best to map the extended data (e.g. emergency voltage bounds) from PTI files into the PowerModels data structure.

ccoffrin:
This sounds good for the near term.  Lets add an optional flag to `parse_file` for keeping all of the extra fields using the PTI names in lower case.

ccoffrin:
@pseudocubic, I say this is done, agreed?

pseudocubic:
Yes, I agree this is done for now. New issues can always opened when we want to make changes like we discussed last week with key renames

PR #260, opened 2018-04-23
ADD: Data extensions from PTI files

pseudocubic:
Adds optional keyword argument `import_all` to `parse_file` (defaults to
false). If true, this option will indicate to the parser that all unused
fields from the PTI file should be passed to the final PowerModels data
structure. PTI keys will be converted to lowercase in the PM data
structure. Adds helper function `import_remaining!(data_out, data_in;
exclude)`.

Adds unit tests to ensure proper passing of additional parameters, and
to check that data structures with additional parameters are still
consistent with previous OPF results.

Updated documentation and changelog.

Addresses #242

### DIFF
--- a/src/io/psse.jl
+++ b/src/io/psse.jl
@@ -107,10 +107,12 @@ end
 
 
 "Imports remaining keys from `data_in` into `data_out`, excluding keys in `exclude`"
-function import_remaining!(data_out::Dict, data_in::Dict; exclude=[])
-    for (k, v) in data_in
-        if k ∉ exclude
-            data_out[lowercase(k)] = v
+function import_remaining!(data_out::Dict, data_in::Dict, import_all::Bool; exclude=[])
+    if import_all
+        for (k, v) in data_in
+            if !(k in exclude)
+                data_out[lowercase(k)] = v
+            end
         end
     end
 end
@@ -134,7 +136,7 @@ function psse2pm_branch!(pm_data::Dict, pti_data::Dict, import_all::Bool)
             sub_data["g_fr"] = pop!(branch, "GI")
             sub_data["b_fr"] = branch["BI"] == 0. && branch["B"] != 0. ? branch["B"] / 2 : pop!(branch, "BI")
             sub_data["g_to"] = pop!(branch, "GJ")
-            sub_data["b_to"] = branch["BJ"] == 0. && branch["B"] != 0. ? pop!(branch, "B") / 2 : pop!(branch, "BJ")
+            sub_data["b_to"] = branch["BJ"] == 0. && branch["B"] != 0. ? branch["B"] / 2 : pop!(branch, "BJ")
             sub_data["rate_a"] = pop!(branch, "RATEA")
             sub_data["rate_b"] = pop!(branch, "RATEB")
             sub_data["rate_c"] = pop!(branch, "RATEC")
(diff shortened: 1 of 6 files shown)


src/io/psse.jl
```diff
@@ -106,36 +106,50 @@ function create_starbus_from_transformer(pm_data::Dict, transformer::Dict)::Dict
...
             sub_data["angmin"] = 0.0
             sub_data["angmax"] = 0.0
             sub_data["transformer"] = false
             sub_data["index"] = i
 
+            if import_all
+                import_remaining!(sub_data, branch)
+            end
+
             append!(pm_data["branch"], [deepcopy(sub_data)])
```
ccoffrin: Do we need to do a deepcopy here?  Also, I think append will still work if this is not wrapped in an array.  Same suggestion for the same block in other functions.

src/io/psse.jl
```diff
@@ -106,36 +106,50 @@ function create_starbus_from_transformer(pm_data::Dict, transformer::Dict)::Dict
 end
 
 
+"Imports remaining keys from `data_in` into `data_out`, excluding keys in `exclude`"
+function import_remaining!(data_out::Dict, data_in::Dict; exclude=[])
+    for (k, v) in data_in
+        if k ∉ exclude
```
ccoffrin: Call me old fashioned, but I try to avoid using non-ASCII characters.  Primarily because I am not 100% sure how well I can debug unicode outside the REPL, especially when doing terminal-based remote development (e.g. using VIM).

src/io/psse.jl
```diff
@@ -501,25 +552,30 @@ end
...
 
-    psse2pm_bus!(pm_data, pti_data)
-    psse2pm_load!(pm_data, pti_data)
-    psse2pm_shunt!(pm_data, pti_data)
-    psse2pm_generator!(pm_data, pti_data)
-    psse2pm_branch!(pm_data, pti_data)
-    psse2pm_transformer!(pm_data, pti_data)
-    psse2pm_dcline!(pm_data, pti_data)
+    if import_all
+        import_remaining!(pm_data, pti_data["CASE IDENTIFICATION"][1]; exclude=["REV"])
```
ccoffrin: why exclude rev?

test/psse.jl
```diff
@@ -319,4 +319,18 @@ end
...
         end
     end
+
+    @testset "import all" begin
+        data = PowerModels.parse_file("../test/data/pti/case30.raw"; import_all=true)
+
+        @test length(data) == 19
+        for (key, n) in zip(["bus", "load", "shunt", "gen", "branch", "dcline"], [30, 21, 2, 6, 41, 0])
+            @test length(data[key]) == n
+        end
```
ccoffrin: I would recommend adding 2-3 spot-checks that specific extended parameter names made it into the data dict of each component.

src/io/psse.jl
pseudocubic: rev is already used above, but pop! didn't work correctly inside the string

src/io/psse.jl
pseudocubic: deepcopy I used because everything is done in a loop, reusing the same sub_data, and i just wanted to be sure i wasn't passing a pointer, which I believe we discussed at one point early on. if you feel confident that its not needed, its no problem to take it out. `append!` does require the second item to be a collection, but in checking this i realize i could use `push!`

src/io/psse.jl
pseudocubic: That's fine with me, I just was not sure how to do this without using `\notin`, what is the correct syntax?

ccoffrin: `!(k in exclude)` will work.

src/io/psse.jl
ccoffrin: aaa yes, I am most often using `push!`

src/io/psse.jl
ccoffrin: To be consistent with the `pop!` behavior with the other parameters consider saving `pop!(pti_data["CASE IDENTIFICATION"][1], "REV")` to a temp variable and then turning it into a version number.

test/psse.jl
pseudocubic: ah, so what i meant to do was check the number of keys per item, like number of keys in each bus, but i made a mistake when i set up the test

ccoffrin:
Looks good.

PR #264, opened 2018-04-25
ADD: Support for HVDC lines and unique "pti_id"s from PTI files

### LINKED ISSUE #212: PSSE dict to PowerModels dict

### LINKED ISSUE #213: Data Formats Documentation

### LINKED ISSUE #231: Add unit tests for PSSE features

pseudocubic:
Adds support for Voltage Source Converters and Two-Terminal DC lines
parsed from PTI files by approximating them via a simple DC line model
with two generators and line losses. For Two-Terminal DC lines, `loss0`
and `loss1` are currently assumed to be zero.

Adds `"source_id"` to supported components imported from PTI files.
`"source_id"` is an Array of `String` and `Int` in each component Dict, the
specifics of which are detailed in the doc-strings of the helper functions
that import each type of component, e.g. `psse2pm_transformer!`.

Adds unit tests for new features to check for OPF and PF solver
feasibility. Added unit tests to check that each supported component
has a `"source_id"` of the correct type.

Updates documentation to indicate supported features of PTI parser. 
`warn` is changed to `info` in several places of the PTI parser. Adds
new features to changelog.

Updates ability of data extensions helper function `import_remaining!`
by allowing for the addition of unsupported component types from PTI
files, and adds unit tests.

Addresses #213 

Closes #212
Closes #231


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ PowerModels.jl Change Log
 - Added support for unit conversion in transformer entries of PTI files
 - Added support for transformer magnetizing admittance from PTI files
 - Added support for extending PowerModels data with all PTI data fields
+- Added support for Voltage Source Converters and Two-Terminal DC lines in PTI files
+- Added "pti_id" to uniquely identify each component imported from a PTI file
```
ccoffrin: On second thought, what do you think about calling this `source_id`?  I am thinking it would be more consistent with `source_type` and `source_version`, and we can make it a standard field that other translators use.

src/io/psse.jl
```diff
@@ -473,49 +513,67 @@ function psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+
+            # How to properly use firing angles to calculate qmin, qmax?
+            # sub_data["qminf"], sub_data["qmaxf"] = calc_2term_reactive_power(power_demand, pop!(dcline, "ANMNR"), pop!(dcline, "ANMXR"))
+            # sub_data["qmint"], sub_data["qmaxt"] = calc_2term_reactive_power(power_demand, pop!(dcline, "ANMNI"), pop!(dcline, "ANMXI"))
+
+            # Relaxed qmin,qmax
+            sub_data["qminf"] = min(sub_data["pminf"], sub_data["pmaxf"])
+            sub_data["qmaxf"] = max(sub_data["pminf"], sub_data["pmaxf"])
+            sub_data["qmint"] = min(sub_data["pmint"], sub_data["pmaxt"])
+            sub_data["qmaxt"] = max(sub_data["pmint"], sub_data["pmaxt"])
```
ccoffrin: Typically `pmin` and `pmax` are positive values, while `qmin`/`qmax` are negative and positive respectively, on a similar order of magnitude of `pmax`.  So I would expect `qmin` to be along the lines of `-max(abs(pmin), abs(pmax))` or `qmin = -qmax`

src/io/psse.jl
```diff
@@ -525,15 +583,62 @@ function psse2pm_dcline!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+            sub_data["br_status"] = pop!(dcline, "MDC") == 0 || pop!(dcside, "TYPE") == 0 || pop!(acside, "TYPE") == 0 ? 0 : 1
+
+            sub_data["pf"] = 0.0
+            sub_data["pt"] = 0.0
+
+            sub_data["qf"] = 0.0
+            sub_data["qt"] = 0.0
+
+            sub_data["vf"] = pop!(dcside, "MODE") == 1 ? pop!(dcside, "ACSET") : 0.0
+            sub_data["vt"] = pop!(acside, "MODE") == 1 ? pop!(acside, "ACSET") : 0.0
```
ccoffrin: default values on these should be 1.0

CHANGELOG.md
pseudocubic: I agree, `source_id` is a better name for this field to make it generalized to any filetype

src/io/psse.jl
pseudocubic: Since you say `pmin`/`pmax` are generally positive, do you think I should also change `pmin` in voltage source converters on line 615-616 to make the minimum zero, instead of `-pmax`?

ccoffrin: VSC looks good to me for now.  We are implementing it as a bi-directional component.

If I recall correctly, in this case we are assuming it is one directional based on the input data?  So on one side the p range will be (0,x) and it will be (-x,0) on the other side.  In either case the q range can be (-x,x) on both sides.

PR #282, opened 2018-05-10
FIX: DCLINE interpretation of HVDC data from PTI files

pseudocubic:
Changes `qmin`, `qmax` -> `-pmax * cos(min_firing_angle) < q < 0`, 
as per discussions with @hakanergun and @frederikgeth, based on
Ch. 10 in "Power System Stability and Control" by Kundur. Removed
unneeded function `calc_2term_reactive_power` and associated
references in documentation.

Fixes `set_costs!` in `test/psse.jl` to change `costs -> [0., 0., 0.]`
for dclines. Also, moved dclines from `IMPEDANCE CORRECTION` to
`VOLTAGE SOURCE CONVERTER` section in `test/data/pti/case7_tplgy.raw`;
was in wrong section. Interesting side note, since data types in the
impedance correction section are nearly all floats, parsing did not fail
despite being in the wrong format.


pseudocubic:
@cjc updated to include warning and check if `-90 < ang < 90`

Issue #278, opened 2018-05-09
Discrepancy in run_opf and run_opf_bf


ccoffrin:
Need to check why `run_opf / SOCWRPowerModel` and `run_opf_bf / SOCDFPowerModel` are producing different results on the `data/matpower/case5_gap.m` test case.  Guessing that this is related to the 4D cut constraint.

@frederikgeth, I won't have time to look into this for a while.  A PR is welcome.

frederikgeth:
Interesting, hopefully I can take a look a week or so.

frederikgeth:
Summarizing the current implementation, you might still get different results because BIM calls 'cut_complex_product_and_angle_difference' whereas BFM does not.

ccoffrin:
This issue is now complete, thanks!  The final point is captured by #244.

PR #286, opened 2018-05-21
Discrepancy BIM vs BFM

frederikgeth:
Regarding #278 , I spent a bit of time going through the implementation. I've found one bug in the bounds of the series active and reactive power flow variables, but that wasn't the source of the discrepancy, as it occurs just as well with the line shunts ('b') set to zero. 

The actual reason is the lack of a current constraint in the SOC BIM formulation. 

In the case5_gap.m, the generator costs are negative. So the objective actually results in a *maximization* of the amount of power injected into the system. As expected, this leads to a significant gap between the exact model and the SOC model, as the slack introduced by the convexification can be used to generate artificial power losses on the lines. Assume the shunts are set to zero, then P_lij + P_lji = P^loss_l = r_l *|I_lij|^2 and (P_lij^2 + Q_lij^2)/|U_i|^2 <= |I_ij|^2. In the BFM, the constraint P_lij + P_lji  = r * ccm limits the amount of power that can be dumped into the branch as an artifical load, due to the bound on ccm (which is nevertheless less strict than the apparent power bound, after compensation for voltage deviations). You can see that all the ccm variables are at their upper bound in case5_gap.m. 

How do you want to go forward?

### DIFF
--- a/docs/src/quickguide.md
+++ b/docs/src/quickguide.md
@@ -86,15 +86,15 @@ This network data can be modified in the same way as the previous Matpower `.m` 
 ## Inspecting AC and DC branch flow results
 The flow AC and DC branch results are not written to the result by default. To inspect the flow results, pass a settings Dict
 ```julia
-result = run_opf("case3_dc.m", ACPPowerModel, IpoptSolver(), setting = Dict("output" => Dict("line_flows" => true)))
+result = run_opf("case3_dc.m", ACPPowerModel, IpoptSolver(), setting = Dict("output" => Dict("branch_flows" => true)))
 result["solution"]["dcline"]["1"]
 result["solution"]["branch"]["2"]
 ```
 
-The losses of a AC or DC branch can be derived:
+The losses of an AC or DC branch can be derived:
 ```julia
-loss_ac =  Dict(name => data["p_to"]+data["p_from"] for (name, data) in result["solution"]["branch"])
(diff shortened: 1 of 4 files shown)


ccoffrin:
I think the natural thing to do is work to the goal of having BIM and BFM defining the same convex set. Thus, I recommend working we make sure to compute a ccm bounds such that are always redundant, namely those bounds will never be a binding constraint.

frederikgeth:
Ok, I updated the bounds to make them redundant in the general case (without assuming any signs for g_sh_fr, b_sh_fr, g_sh_to, b_sh_to, r or x). The objective values for BFM and BIM for case5_gap are now and -28238.0 and -28237.3, which I assume is within the margin of accuracy?

ccoffrin:
In my experience even those small differences are most often caused by slight deviations in the mathematical formulations.  On this particular test case, is it possible that the 0.7 difference is due to the 4D cuts in the BIM model?  If not, I would look another subtle difference in the formulation or bounds.

frederikgeth:
Just tested this hypothesis. Disabling the call to cut_complex_product_and_angle_difference for BIM indeed leads to an objective value of -28238.0.

ccoffrin:
Perfect.

frederikgeth:
A follow-up with a very long delay - see https://arxiv.org/abs/2202.06449 - closing the loop for whomever ends up here in the future.

PR #289, opened 2018-05-24
FIX: DistFlow formulation for multiphase

pseudocubic:
adds `getmpv` functions to `constraint_voltage_angle_difference` to
handle multiphase data

Adds `getmpv` formulations.

### DIFF
--- a/src/core/multiphase.jl
+++ b/src/core/multiphase.jl
@@ -105,5 +105,7 @@ end
 
 
 getmpv(value::Any, phase::Int) = value
+getmpv(value::Any, phase_i::Int, phase_j::Int) = value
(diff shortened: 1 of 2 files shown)


ccoffrin:
The root cause of the issue was that this constraint is not following the PowerModels standard style.  The convention is that all of the data extractions happen in the constraint template, before getting to an actual constraint implementation.  

Would you like to address that point now?  Or make an issue to resolve it down the line?

pseudocubic:
I think make an issue and address it later, just to keep things moving; changing it properly would touch a lot of functions, slowing me down on my other priorities.

ccoffrin:
Issue made here, https://github.com/lanl-ansi/PowerModels.jl/issues/291

PR #293, opened 2018-05-30
ADD: extension to `Base.var`

pseudocubic:
Adds `var(pm, nw)` for use in SOC formulation in ThreePhasePowerModels


ccoffrin:
For consistency with `ref` and `var` should we add similar functions to `con` as well?

pseudocubic:
Ah, yes, probably; I hadn't looked closely at those functions before.

PR #308, opened 2018-06-17
Revised Multiphase Objective Functions

ccoffrin:
@pseudocubic, @bluejuniper after discussing with @hakanergun at PSCC, we agreed the best near term semantics for multi-phase generator cost functions is to have one function per generator and sum active power over the phases before applying the function.  As per this PR.  Once once this is merged ThreePhasePowerModels will need to be updated, I'll leave that to @pseudocubic.

@harshangrjn and @hhijazi, please let us know if this does not match what you had in mind.

CC @rb004f, @frederikgeth

### DIFF
--- a/src/core/data.jl
+++ b/src/core/data.jl
@@ -234,7 +234,7 @@ function _make_per_unit(data::Dict{String,Any}, mva_base::Real)
         apply_func(dcline, "qmaxf", rescale)
         apply_func(dcline, "qminf", rescale)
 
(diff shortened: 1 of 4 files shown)


ccoffrin:
Great.  You are correct, we have indeed eliminated the need for `getmpv` for now; but I still think it's good to have it around if a similar case arises in the future.  Have array data for each phase seems like it could come up from time to time.

ccoffrin:
No complaints, so I am merging this.

PR #309, opened 2018-06-20
Matpower export

rb004f:
@ccoffrin First cut at an matpower export. Take a look


src/io/matpower.jl
```diff
@@ -504,8 +506,8 @@ function mp2pm_dcline(data::Dict{String,Any})
         loss0 = dcline["loss0"]
         loss1 = dcline["loss1"]
 
-        delete!(dcline, "pmin")
-        delete!(dcline, "pmax")
+#        delete!(dcline, "pmin")
+#        delete!(dcline, "pmax")
```
ccoffrin: I think leaving these in will risk confusion on which parameter PowerModels is actually using.  I recommend renaming to `mp_pmin` and `mp_pmax` to mitigate confusion.

src/io/matpower.jl
```diff
@@ -665,3 +667,419 @@ function merge_generic_data(data::Dict{String,Any})
     end
 end
 
+"Export a power flow case in matpower format"
+function export_matpower(io::IO, data::Dict{String,Any})
+
```
ccoffrin: In stead of doing the mixed-unit conversion while writing the output, can you just call `make_mixed_units(data)` at this point?

src/io/matpower.jl
```diff
@@ -665,3 +667,419 @@ function merge_generic_data(data::Dict{String,Any})
...
+                    "\t", (gen["pg"]*mvabase),
+                    "\t", (gen["qg"]*mvabase),
+                    "\t", (gen["qmax"]*mvabase),
+                    "\t", (gen["qmin"]*mvabase),
+                    "\t", gen["vg"],
+                    "\t", gen["mbase"],
+                    "\t", gen["gen_status"],
+                    "\t", (gen["pmax"]*mvabase),
+                    "\t", (gen["pmin"]*mvabase),
+                    "\t", gen["pc1"],
```
ccoffrin: By the technical Matpower spec I believe this is where the `haskey` checks should start.

src/io/matpower.jl
```diff
@@ -665,3 +667,419 @@ function merge_generic_data(data::Dict{String,Any})
...
+    i = 1
+    for (idx,branch) in sort(branches)
+        if idx != branch["index"]
+            warn(LOGGER, "The index of the branch does not match the matpower assigned index. Any data that uses branch indexes for reference is corrupted.");           
+        end 
+        println(io, "\t", branch["f_bus"], 
+                    "\t", branch["t_bus"], 
+                    "\t", branch["br_r"], 
+                    "\t", branch["br_x"], 
+                    "\t", (haskey(branch,"b_to") ? branch["b_to"] + branch["b_fr"]  : 0),
```
ccoffrin: I don't believe `b_to` is not optional, I think you can hard code this as `branch["b_to"] + branch["b_fr"]`

src/io/matpower.jl
```diff
@@ -665,3 +667,419 @@ function merge_generic_data(data::Dict{String,Any})
...
+    
+    # Print the dcline cost data
+    if length(data["dcline"]) > 0
+        if haskey(collect(values(dclines))[1], "cost")
+            println(io, "%% dcline cost data")
+            println(io, "%    1    startup    shutdown    n    x1    y1    ...    xn    yn")
+            println(io, "%    2    startup    shutdown    n    c(n-1)    ...    c0")
+            println(io, "mpc.dclinecost = [")
+        
+            for (idx,dcline) in (sort(dclines))
```
ccoffrin: This block will always be identical to the generator one, might want to make it a function.

test/data/matpower/extra_components.m
```diff
@@ -0,0 +1,67 @@
+%% MATPOWER Case Format : Version 2
+mpc.version = '2';
```
ccoffrin: case3.m is nominally the place to test the extended Matpower files.  Can you extend case3.m to cover your cases instead of adding a new case?

test/export.jl
```diff
@@ -0,0 +1,97 @@
...
+function test_case(filename::AbstractString)
+    temp_file = "temp.m"
+    source_data = PowerModels.parse_file(filename)        
+
+    io = open(temp_file, "w")         
+    PowerModels.export_matpower(io, source_data)
+    close(io)    
+    destination_data = PowerModels.parse_file(temp_file)  
+    
+    @test InfrastructureModels.compare_dict(source_data, destination_data) == true
```
ccoffrin: I think you can omit the `== true`

test/export.jl
```diff
@@ -0,0 +1,97 @@
+
```
ccoffrin: These tests would fit well in `test/data.jl`

rb004f:
@ccoffrin pull request now addresses your comments. I will note that by changing case3.m, I had to change two other tests.  You may want to double check them.

PR #312, opened 2018-06-24
Remove Series Variables from Branch Flow Formulation

ccoffrin:
@frederikgeth, I updated the formulation to not require the series variables and updated a few items to be consistent with the branch flow naming.  Any concerns about dropping the explicit series representation?


frederikgeth:
No objections. 

Wondering though how you see the type hierarchy evolve though. 

For the SOC BFM derived from the DistFlow formulation it is:
```
SOCBFForm <: AbstractDFForm <: AbstractPowerFormulation end
SOCBFPowerModel = GenericPowerModel{SOCBFForm}
```
For the SOC BIM derived it is:
```
SOCWRForm <: AbstractWRForm <: AbstractPowerFormulation end
SOCWRPowerModel = GenericPowerModel{SOCWRForm}
```

Their intersection is:
`AbstractWForms = Union{AbstractWRForms, AbstractDFForm}`

So the BIM formulations (concrete and abstract) are named after the variable space, but the concrete BFM after the BFM itself, whereas the abstract still refers to DistFlow.

ccoffrin:
Good catch, for consistency I updated `AbstractDFForm` to `AbstractBFForm` for now.  

Regarding the type hierarchy plans; You make a good point, in the long run I would like to merge `SOCWRForm` and `SOCBFForm` into a common type (which is along the lines of a generic SOC model) and also add support for some non-convex models in `opf_bf` (e.g. ACP).  But this will take some thought, as you know there are many formulation alternatives and it is not quite clear if there is an easy way they can share types.

frederikgeth:
Ok, clear for now. A colleague of mine may be interested in implementing the nonconvex DistFlow formulation, we'll keep you posted.

PR #323, opened 2018-07-09
FIX: Line splitting in PTI parser

pseudocubic:
In String fields, like "name", there can be commas, which was causing
special cases (e.g. transformers) to split in the wrong places. Switched
all cases to use `get_line_elements` function, which relies on RegEx
and handles this case where a comma is in a String field surrounded by
single quotes.


ccoffrin:
Looks good to me.

PR #329, opened 2018-07-11
FIX: Single-quote in busnames

### LINKED ISSUE #327: single-quote characters inside of bus names
There can apparently be single quotations inside of names in PTI files, e.g. `'10' DROP    '`. Current method of using regex cannot handle this.

pseudocubic:
Fixes problem where there could be a quotation mark inside of a bus
name by taking advantage of the fact that all busnames in PTI v33.3
must be exactly 12 characters long to be valid.

Closes #327


ccoffrin:
You should update one of the test cases to exercise this bug, so that we don't have a regression in the future.

PR #333, opened 2018-07-18
SDP relaxation with constraint decomposition

kersulis:
This PR adds functionality for solving the SDP relaxation with decomposition of the semidefinite constraint. Replacing one large semidefinite constraint with a collection of small constraints (each associated with a maximal clique of a chordal extension of the system graph) can dramatically reduce solver times. See [here](https://molzahn.github.io/pubs/kersulis_hiskens_coffrin_molzahn-graph_metrics2018.pdf) for details. The application of this technique to the SDP OPF problem is due to Dan Molzahn, whose MATLAB implementation I have been comparing against.

Next steps:
- [x] Ongoing: quantify improvement by comparing MATLAB, PowerModels.jl without decomposition, and PowerModels.jl with decomposition. (Use Mosek for consistency with MATLAB code.) @ccoffrin how do you efficiently generate that big OPF results table? pandas has markdown table export, but I don't see that in DataFrames.jl.
- [x] Compare SCS and Mosek (results so far suggest SCS may not solve large instances in reasonable times even with decomposition).
- [ ] Add clique merge functionality (application of this technique in SDP OPF setting also due to Dan Molzahn) to further reduce solver time. (currently debugging)
- [ ] Fix infeasible cases
  - [x] Infeasibility due to negative generator active power outputs kersulis#1
  - [ ] Infeasibility due to multiple network components (e.g. case3375)
- [ ] Add objective function scaling to improve numerics, #368


ccoffrin:
I don't see where you add the equality constraints for variables that are shared across the groups.

Also, feel free to put the implementation in `wrm.jl`

kersulis:
The WR and WI variables are generated the same as in SDPWRMForm, but in `constraint_voltage`  I generate overlapping semidefinite constraints rather than one large constraint. Because I am just constraining groups of the same WR and WI variables, shouldn't the linking constraints be implicit? i.e. the solver should know there can be only one value for each variable, regardless of how many semidefinite constraints it shows up in.

Here is Mosek's description of the `case30` OPF SDP relaxation problem without decomposition (and before presolve):
```
Constraints            : 1211            
Cones                  : 45              
Scalar variables       : 531             
Matrix variables       : 1               
Integer variables      : 0           
```
And with decomposition (again, before presolve):
```
Constraints            : 953             
Cones                  : 45              
Scalar variables       : 531             
Matrix variables       : 12              
Integer variables      : 0               
```
There may be some benefit to making the linking constraints explicit, but the results I've observed so far are consistent with those of Dan's MATLAB implementation: `SDPDecompForm` reaches the same objective value as `SDPWRMForm` in roughly half the time (0.19s vs 0.07s for case30, for example).

When I use SCS with `SDPDecompForm`, I do get messages like this for some cases (case6 for example):
```
WARN: A->p (column pointers) not strictly increasing, column 1 empty
WARN: A->p (column pointers) not strictly increasing, column 2 empty
WARN: A->p (column pointers) not strictly increasing, column 3 empty
WARN: A->p (column pointers) not strictly increasing, column 5 empty
WARN: A->p (column pointers) not strictly increasing, column 8 empty
WARN: A->p (column pointers) not strictly increasing, column 9 empty
WARN: A->p (column pointers) not strictly increasing, column 13 empty
WARN: A->p (column pointers) not strictly increasing, column 17 empty
WARN: A->p (column pointers) not strictly increasing, column 19 empty
```
Despite these messages, SCS still reaches the same objective value, and faster than without decomposition (though the speedup is less for small networks like case6).

kersulis:
This new implementation creates only those voltage product variables that are needed in the decomposed semidefinite constraints. It also makes the linking constraints explicit.

Here is a comparison of Mosek solver output for `nesta_case30_ieee`:

|                   | SDPDecompForm | SDPWRMForm |
|:------------------|:--------------|:-----------|
| Constraints       | 2261          | 3949       |
| Cones             | 88            | 88         |
| Scalar variables  | 834           | 1811       |
| Matrix variables  | 26            | 1          |
| Integer variables | 0             | 0          |
| Objective         | 204.96827     | 204.96835  |
| Solver time       | 0.11          | 2.90       |

kersulis:
@molzahn, @bghaddar

kersulis:
After the cleanup, only `SDPDecompForm` remains. This implementation does not do clique merge. The only heuristic used is the sparsity-preserving Cholesky factorization that yields a chordal graph extension. I do not currently know of a better way to obtain a chordal extension, but I'll let you know if I find one :)

src/form/wrm.jl
```diff
@@ -1,21 +1,38 @@
...
+end
+
 ""
 SDPWRMPowerModel(data::Dict{String,Any}; kwargs...) = GenericPowerModel(data, SDPWRMForm; kwargs...)
 
+SDPDecompPowerModel(data::Dict{String,Any}; kwargs...) = GenericPowerModel(data, SDPDecompForm; kwargs...)
 
 ""
-function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true) where T <: AbstractWRMForm
+function variable_voltage(pm::GenericPowerModel{SDPWRMForm}; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true)
```
ccoffrin: This function signature should be, 
```
function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true) where T <: SDPWRMForm
```
Otherwise it cannot be extended later on.

src/form/wrm.jl
```diff
@@ -76,19 +93,340 @@ function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.
         var(pm, nw, cnd, :wr)[(i,j)] = WR[w_fr_index, w_to_index]
         var(pm, nw, cnd, :wi)[(i,j)] = WI[w_fr_index, w_to_index]
     end
-
 end
 
+function variable_voltage(pm::GenericPowerModel{SDPDecompForm}; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded=true)
```
ccoffrin: This function signature should be,
```
function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true) where T <: SDPDecompForm
```
Otherwise it cannot be extended later on.

src/form/wrm.jl
```diff
@@ -76,19 +93,340 @@ function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.
...
+                    var(pm, nw, cnd, :wi)[(i_bus, j_bus)] = WI[i, j]
+                end
+            end
+        end
+    end
+end
 
 ""
-function constraint_voltage(pm::GenericPowerModel{T}, nw::Int, cnd::Int) where T <: AbstractWRMForm
+function constraint_voltage(pm::GenericPowerModel{SDPWRMForm}, nw::Int, cnd::Int)
```
ccoffrin: Same as above.

src/form/wrm.jl
```diff
@@ -76,19 +93,340 @@ function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.
...
     WR = var(pm, nw, cnd)[:WR]
     WI = var(pm, nw, cnd)[:WI]
 
     @SDconstraint(pm.model, [WR WI; -WI WR] >= 0)
+end
+
+function constraint_voltage(pm::GenericPowerModel{SDPDecompForm}, nw::Int, cnd::Int)
```
ccoffrin: Same as above.

src/form/wrm.jl
```diff
@@ -76,19 +93,340 @@ function variable_voltage(pm::GenericPowerModel{T}; nw::Int=pm.cnw, cnd::Int=pm.
...
+        # Lower-dimensional SOC constraint equiv. to SDP for 2-vertex
+        # clique
+        if ng == 2
+            wr_ii = WR[1, 1]
+            wr_jj = WR[2, 2]
+            wr_ij = WR[1, 2]
+            wi_ij = WI[1, 2]
+
+            # standard SOC form (Mosek doesn't like rotated form)
+            @constraint(pm.model, (wr_ii + wr_jj)/2 >= norm([(wr_ii - wr_jj)/2; wr_ij; wi_ij]))
```
ccoffrin: I think this constraint is off by a factor of 2.  Double check the math.  In either case, this version should work,
```
@constraint(pm.model, (wr_ii + wr_jj) >= norm([(wr_ii - wr_jj); 2*wr_ij; 2*wi_ij]))
```

src/form/wrm.jl
kersulis: Good point. You mean `AbstractWRMForm`, right?

src/form/wrm.jl
kersulis: @molzahn: We checked this against the derivation yesterday, right?

ccoffrin:
It's up to you. Currently we don't have clear conventions on how to name these types, so I won't be too picky about the name at this point.

You might consider something like SDPWRMCFForm (CF is short for cholfact) or SDPWRMCFCEForm (CFCE is short for cholfact chordal extension).  Try to think of a name that will be less likely to collide with other possible implementations.

Also, following in the new [model documentation convention](https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/acp.jl#L39), consider adding a doc string with your preferred references.

src/form/wrm.jl
```diff
@@ -256,9 +254,11 @@ function constraint_voltage(pm::GenericPowerModel{SDPDecompForm}, nw::Int, cnd::
             wr_jj = WR[2, 2]
             wr_ij = WR[1, 2]
             wi_ij = WI[1, 2]
+            wi_ji = WI[2, 1]
 
             # standard SOC form (Mosek doesn't like rotated form)
-            @constraint(pm.model, (wr_ii + wr_jj)/2 >= norm([(wr_ii - wr_jj)/2; wr_ij; wi_ij]))
+            @constraint(pm.model, (wr_ii + wr_jj) >= norm([(wr_ii - wr_jj); 2*wr_ij; 2*wi_ij]))
+            @constraint(pm.model, wi_ij == -wi_ji)
```
ccoffrin: FYI, in my past experience adding this symmetry constraint was causing some numerical issues in the solvers.

test/opf.jl
```diff
@@ -564,13 +564,13 @@ end
         data = PMs.parse_file("../test/data/matpower/case14.m")
         pm = GenericPowerModel(data, SDPDecompForm)
 
-        cadj, lookup_index = PMs.chordal_extension(pm)
+        cadj, lookup_index, σ = PMs.chordal_extension(pm)
```
ccoffrin: Call me old fashioned but I prefer not to use unicode in PowerModels code.  Just use sigma instead.

PR #335, opened 2018-07-25
OBBT functionality

kaarthiksundar:
Optimality-based bound tightening functionality is added to PowerModels. The entry point for the algorithm is the function ``run_obbt_opf(case, model_constructor, solver)`` which can be found in src/util/obbt.jl

Unit tests using the case3.m are added. More unit tests to come in subsequent PRs.


src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+            <= bound 
+    )
+end 
+
+function constraint_obj_bound(pm::GenericPowerModel, bound)
+    model = PowerModels.check_cost_models(pm)
+    if model != 2 
+        error("Only cost models of type 2 is supported at this time, given cost model type $(model)")
+    end 
+    constraint_obj_bound_polynomial(pm, bound)
```
ccoffrin: No need to break this out as a separate function.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+    
+    modifications = Dict{String,Any}()
+
+    modifications["per_unit"] = true
+    modifications["bus"] = Dict{String,Any}()
+    modifications["branch"] = Dict{String,Any}()
+
+    buses = ids(pm, :bus)
+    buspairs = ids(pm, :buspairs)
+    branches = ids(pm, :branch)
```
ccoffrin: Seems like there is no need to make these stand alone variables.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+end 
+
+function create_modifications(pm::GenericPowerModel, lb::Float64, ub::Float64, id::Int64)
+    
+    modifications = Dict{String,Any}()
+
+    modifications["per_unit"] = true
+    modifications["bus"] = Dict{String,Any}()
+
+    index = string(ref(pm, :bus, id, "index"))
```
ccoffrin: id and index are the same.  you can do `"$(id)"`

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+        f_bus = ref(pm, :branch, branch, "f_bus") 
+        t_bus = ref(pm, :branch, branch, "t_bus")
+        bp = (f_bus, t_bus)
+        modifications["branch"][index] = Dict{String,Any}( "angmin" => td_lb[bp], "angmax" => td_ub[bp] )
+    end 
+
+    return modifications
+end 
+
+function create_modifications(pm::GenericPowerModel, lb::Float64, ub::Float64, id::Int64)
```
ccoffrin: a more descriptive name would be `create_bus_modification`.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+    modifications["bus"] = Dict{String,Any}()
+
+    index = string(ref(pm, :bus, id, "index"))
+    modifications["bus"][index] = Dict{String,Any}( "vmin" => lb, "vmax" => ub )
+
+    return modifications
+
+end 
+
+function create_modifications(pm::GenericPowerModel, lb::Float64, ub::Float64, id::Tuple{Int64,Int64})
```
ccoffrin: a more descriptive name would be `create_branch_modification`.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+
+    modifications = Dict{String,Any}()
+
+    modifications["per_unit"] = true
+    modifications["branch"] = Dict{String,Any}()
+
+    branches = ids(pm, :branch)
+
+
+    for branch in branches
```
ccoffrin: Putting this loop in hear will result in a `|branches|^2` runtime to update all of the branches.  Given the style of create_modifications for buses, I would recommend that the `id` in this function be the branch id.  The 1-to-many mapping for pairs to branches should be handled in the loop that calls this function.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+    max_iter = 100, 
+    time_limit = 3600.0,
+    upper_bound = Inf,
+    upper_bound_constraint = false, 
+    sequential_obbt = false,
+    rel_gap_tol = Inf,
+    min_bound_width = 1e-2,
+    improvement_tol = 1e-3, 
+    output_tol = 1e+5,
+    termination = "avg",
```
ccoffrin: Should this be a symbol?

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+    return modifications
+
+end
+
+function run_obbt_opf(file::String, model_constructor, solver; kwargs...)
+    data = PowerModels.parse_file(file)
+    return run_obbt_opf(data, model_constructor, solver; kwargs...)
+end 
+
+function run_obbt_opf(data::Dict{String,Any}, model_constructor, solver;
```
ccoffrin: I would recommend having a recommended default model constructor.  I think the average user just wants the bounds, not too much about how they are calculated.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+    current_relaxation_objective = result_relaxation["objective"]
+    if upper_bound < current_relaxation_objective
+        error(LOGGER, "the upper bound provided to OBBT is not a valid ACOPF upper bound")
+    end 
+    if !(result_relaxation["status"] in status_pass) 
+        error(LOGGER, "relaxation is infeasible, implying the ACOPF is infeasible")
+    end 
+    current_rel_gap = Inf 
+    (!isinf(rel_gap_tol)) && (current_rel_gap = (upper_bound - current_relaxation_objective)/upper_bound)
+    info(LOGGER, "Initial relaxation gap = $current_rel_gap")
```
ccoffrin: Maybe only print this if `upper_bound` is specified.

test/runtests.jl
```diff
@@ -52,6 +52,8 @@ include("common.jl")
 
     include("multi-nw-cnd.jl")
 
+    include("bt.jl")
```
ccoffrin: Call this file `util` or `obbt`, so it matches the other file structures.

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+function run_obbt_opf(data::Dict{String,Any}, model_constructor, solver;
+    max_iter = 100, 
+    time_limit = 3600.0,
+    upper_bound = Inf,
+    upper_bound_constraint = false, 
+    sequential_obbt = false,
+    rel_gap_tol = Inf,
+    min_bound_width = 1e-2,
+    improvement_tol = 1e-3, 
+    output_tol = 1e+5,
```
ccoffrin: Wondering if this is too generous given a solver's typical bounds?

src/util/obbt.jl
```diff
@@ -0,0 +1,406 @@
...
+            end
+            total_td_reduction += (td_reduction)
+            max_td_reduction = max(td_reduction, max_td_reduction)
+
+        end 
+        avg_td_reduction = total_td_reduction/length(buspairs)
+
+        td_range_final = sum([td_ub[bp] - td_lb[bp] for bp in buspairs])
+        
+        info(LOGGER, "iteration $(current_iteration+1), vm range: $vm_range_final, td range: $td_range_final")
```
ccoffrin: Would the objective value of the current relaxation may also be of interest?

src/util/obbt.jl
kaarthiksundar: converted to a single function

src/util/obbt.jl
kaarthiksundar: Agreed, removed.

src/util/obbt.jl
kaarthiksundar: removed in lieu of sequential_obbt removal

src/util/obbt.jl
kaarthiksundar: removed in lieu of sequential_obbt removal

src/util/obbt.jl
kaarthiksundar: removed in lieu of sequential_obbt removal

src/util/obbt.jl
kaarthiksundar: removed in lieu of sequential_obbt removal

src/util/obbt.jl
kaarthiksundar: done, model_constructor is now a keyword argument

src/util/obbt.jl
kaarthiksundar: On larger instances, setting an output_tol of 1e+4 results in increased bounds. Nevertheless, the bounds are affected only in the second decimal place. Hence, I am making the default to 1e+4. Also, this option is being changed to precision.

src/util/obbt.jl
kaarthiksundar: agreed, converted to a symbol

src/util/obbt.jl
kaarthiksundar: done

src/util/obbt.jl
kaarthiksundar: agreed, logging the objective value in the updated PR

test/runtests.jl
kaarthiksundar: changed to util.jl

ccoffrin:
@kaarthiksundar is this more or less ready to go?  One optional thing to consider would be adding a page to the docs.  I would recommend a "Utilities" page under the Manual section, where one of the subsections was about OBBT.

kaarthiksundar:
Yes it is ready. I shall get the docs done. Then I think it should be ready for the merge.

ccoffrin:
Great.  I am working on the issue with Pajarito and running tests.

src/util/obbt.jl
```diff
@@ -191,23 +193,30 @@ function run_obbt_opf(data::Dict{String,Any}, solver;
         for bus in buses
             (vm_ub[bus] - vm_lb[bus] < min_bound_width) && (continue)
             
+            start_time = time()
             # vm lower bound solve 
             lb = NaN 
             @objective(model_bt.model, Min, vm[bus])
             result_bt = solve_generic_model(model_bt, solver)
-            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal)
+            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal || result_bt["status"] == :SubOptimal)
```
ccoffrin: In my experince, SubOptimal status from Gurobi means that the solution and objective value are not correct.  This should be treated more like `:Infeasible` than `:Optimal`.

src/util/obbt.jl
```diff
@@ -191,23 +193,30 @@ function run_obbt_opf(data::Dict{String,Any}, solver;
...
+            else 
+                error(LOGGER, "BT minimization problem for vm[$bus] errored - change tolerances.")
             end
 
             #vm upper bound solve
             ub = NaN 
             @objective(model_bt.model, Max, vm[bus])
             result_bt = solve_generic_model(model_bt, solver)
-            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal)
+            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal || result_bt["status"] == :SubOptimal)
```
ccoffrin: Same as above.

src/util/obbt.jl
```diff
@@ -243,23 +252,30 @@ function run_obbt_opf(data::Dict{String,Any}, solver;
         for bp in buspairs
             (td_ub[bp] - td_lb[bp] < min_bound_width) && (continue)
 
+            start_time = time()
             # td lower bound solve 
             lb = NaN 
             @objective(model_bt.model, Min, td[bp])
             result_bt = solve_generic_model(model_bt, solver)
-            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal)
+            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal || result_bt["status"] == :SubOptimal)
```
ccoffrin: Same as above.

src/util/obbt.jl
```diff
@@ -243,23 +252,30 @@ function run_obbt_opf(data::Dict{String,Any}, solver;
...
+            else
+                error(LOGGER, "BT minimization problem for td[$bp] errored - change tolerances")
             end
 
             # td upper bound solve 
             ub = NaN 
             @objective(model_bt.model, Max, td[bp])
             result_bt = solve_generic_model(model_bt, solver)
-            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal)
+            if (result_bt["status"] == :LocalOptimal || result_bt["status"] == :Optimal || result_bt["status"] == :SubOptimal)
```
ccoffrin: Same as above.

docs/src/utilities.md
```diff
@@ -0,0 +1,58 @@
+# PowerModels Utility Functions
```
ccoffrin: This looks good, just needs to be added the pages field in `docs/make.jl`

src/util/obbt.jl
ccoffrin: Stop by ANSI conference room if you would like to discuss.

ccoffrin:
Looks good.  Thanks!

PR #336, opened 2018-08-01
FIX: import_remaining array eltype check

pseudocubic:
import_remaining! did not check for element type of arrays components, which broke behavior in ThreePhasePowerModels. Now checks to ensure that arrays have Dict elements.

Closes ThreePhasePowerModels#51


ccoffrin:
If you fix a bug in a downstream package, the best practice is to also add a test in this package that exhibits the original error, so that it does not re-occur during a major refactor / rewrite.

PR #348, opened 2018-08-10
Fix Bug in Quadratic Conic Objectives

ccoffrin:


ccoffrin:
@kersulis, fixed on master.

Issue #343, opened 2018-08-02
Formulation Documentation


ccoffrin:
Some documentation should be added to discuss the core formulations, e.g. ACP, ACR, SOCWR, QCWR, SDPWRM.  Docs should discuss the origin of the model and enhancements that have been added to more accurately support the complete network model (e.g. bus shunts, line charge).

This would address #8 as well.

frederikgeth:
yes!

PR #352, opened 2018-08-14
Updated docs

frederikgeth:
Made some progress towards #343 and #8.

Feel free to comment on the choice of references, @ccoffrin .

Eventually, a figure should be added to explain the standardized line model.

### DIFF
--- a/docs/src/formulation-details.md
+++ b/docs/src/formulation-details.md
@@ -1,6 +1,6 @@
 # Formulation Details
 
-This section provides references to understand the formulations as implemented in PowerModels. The list is not meant as a literature discussion, but to give the main starting points to understand the implementation of the formulations.
+This section provides references to understand the formulations as provided by PowerModels. The list is not meant as a literature discussion, but to give the main starting points to understand the implementation of the formulations.
 
 The history of AC OPF formulations up to 2012 is discussed in:
 - Cain, M. B., O’ Neill, R. P., & Castillo, A. (2012). History of optimal power flow and formulations. <https://www.ferc.gov/industries/electric/indus-act/market-planning/opf-papers/acopf-1-history-formulation-testing.pdf>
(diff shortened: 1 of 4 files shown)


docs/src/formulation-details.md
```diff
@@ -0,0 +1,85 @@
+# Formulation Details
+
```
ccoffrin: We should add this reference, Daniel Molzahn and Ian Hiskens. ”A Survey of Relaxations and Approximations of the Power Flow Equations" Foundations and Trends in Electric Energy Systems, forthcoming.

docs/src/formulation-details.md
```diff
@@ -0,0 +1,85 @@
...
+This section provides references to understand the formulations as provided by PowerModels. The list is not meant as a literature discussion, but to give the main starting points to understand the implementation of the formulations.
+
+The history of AC OPF formulations up to 2012 is discussed in:
+- Cain, M. B., O’ Neill, R. P., & Castillo, A. (2012). History of optimal power flow and formulations. <https://www.ferc.gov/industries/electric/indus-act/market-planning/opf-papers/acopf-1-history-formulation-testing.pdf>
+
+A handbook for developing an understanding of the principles of convex relaxation applied to optimal power flow is:
+- Taylor, J. A. (2015). Convex optimization of power systems (1st ed.). Cambridge University Press.
```
ccoffrin: In place of this reference I would recommend,
- https://arxiv.org/abs/1807.07227
- https://ieeexplore.ieee.org/document/7271127/

docs/src/formulation-details.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Notes on the mathematical model for all formulations
+PowerModels implements a slightly generalized version of the AC Optimal Power Flow problem from [Matpower](http://www.pserc.cornell.edu/matpower/), as discussed [here](./math-model.html).
+
```
ccoffrin: I believe there is a better way to do cross links within the documentation.  Check the Documentor's docs about that.

Also this is a good place to reference https://youtu.be/j7r4onyiNRQ

docs/src/formulation-details.md
```diff
@@ -0,0 +1,85 @@
...
+- A bus defines a complex power balance for all the sets lines, generators, loads, bus shunts connected to it, i.e. one can define multiple load  and shunt  components on each bus $i$
+
+Nevertheless, in the literature, a simplified bus model is often used:
+- Only a single (aggregated) load per bus is supported
+- Only a single (aggregated) bus shunt per bus is supported
+
+
+
+
+## `ACPPowerModel`
```
ccoffrin: I think we can get a more maintainable solution to this by putting these links in the docstring when the type is defined, then autodoc will pick them up automatically. If you want to call out specific doc strings items at this location, there is a way to do that as well.

I would also consider adding a bibtex entry to the docstring.

docs/src/formulation-details.md
frederikgeth: done!

docs/src/formulation-details.md
frederikgeth: done!

docs/src/formulation-details.md
frederikgeth: done

docs/src/formulation-details.md
frederikgeth: Great suggestion, that's a better approach indeed. This will take a bit of time and my agenda has been filled up for the rest of the week, so I'll re-visit later.

ccoffrin: Ok, I'll merge this now and you can add an issue on the update.

Issue #316, opened 2018-06-26
DC line model breaks PWL generator cost format


claytonpbarrows:
e.g. [case3_pwlc.zip](https://github.com/lanl-ansi/PowerModels.jl/files/2135479/case3_pwlc.zip)
the file with the dc line field results in the following error:
```
network_data = PowerModels.parse_file("case3_pwlc_dc.m")
[warn | InfrastructureModels]: Matlab parser skipping the following line:
  ]
[warn | InfrastructureModels]: Matlab parser skipping the following line:
  ]
[warn | PowerModels]: added zero cost function data for dclines
[warn | PowerModels]: added zero cost function data for dclines
[warn | PowerModels]: pwl x value 0.12 is outside the generator bounds 0.0-0.0
[warn | PowerModels]: pwl x value 0.12 is outside the generator bounds 0.0-0.0
[warn | PowerModels]: pwl x value 0.16 is outside the generator bounds 0.0-0.0
[warn | PowerModels]: pwl x value 0.16 is outside the generator bounds 0.0-0.0
[warn | PowerModels]: pwl x value 0.2 is outside the generator bounds 0.0-0.0
[warn | PowerModels]: pwl x value 0.2 is outside the generator bounds 0.0-0.0

non-increasing x values in pwl cost model
```

ccoffrin:
Confirmed this is a bug, thanks for finding it!  Should be fixed in master now.  If you would like to apply the fix locally, only a few lines need to be updated, have a look at https://github.com/lanl-ansi/PowerModels.jl/pull/317.

jd-lara:
@claytonpbarrows is it possible to update the branch `moi-julia-v0.7` to include this fix?

ccoffrin:
Short answer, not for some time.

When Ipopt is woking again with MOI, I will try to update the `moi` branch to PowerModels master.  Then you can look at merging that in `moi-julia-v0.7`.  If you need a quick fix, you can re-do #317 for on `moi-julia-v0.7`.

PR #355, opened 2018-08-15
Jd/moi julia v0.7

jd-lara:
Enables parsing of matpower files from Julia v0.7. It will still fail for the `RTS - GLMC` data since the fix for #316 hasn't been implemented in the MOI branch.


REQUIRE
```diff
@@ -1,6 +1,8 @@
-julia 0.6
+julia 0.7-
 JSON 0.6.0
-JuMP 0.17
+JuMP
```
ccoffrin: min bound should be explicit here.

src/io/matpower.jl
```diff
@@ -557,7 +557,7 @@ function add_dcline_costs(data::Dict{String,Any})
                     "startup" => 0.0,
                     "shutdown" => 0.0,
                     "ncost" => 2,
-                    "cost" => [0.0, 0.0, 0.0, 0.0]
+                    "cost" => [dcline["pminf"], 0.0, dcline["pmaxf"], 0.0]
```
ccoffrin: This fix belongs in PowerModels master.

src/io/matpower.jl
```diff
@@ -546,7 +546,7 @@ end
 
 "adds dcline costs, if gen costs exist"
 function add_dcline_costs(data::Dict{String,Any})
-    if length(data["gencost"]) > 0 && length(data["dclinecost"]) <= 0
+    if length(data["gencost"]) > 0 && length(data["dclinecost"]) <= 0 && length(data["dcline"]) > 0
```
ccoffrin: This fix belongs in PowerModels master.

test/data/matpower/case5_pwlc.m
```diff
@@ -52,3 +52,7 @@
 	4	 10	 0.00297	 0.0297	 0.00674	 240.0	 240.0	 240.0	 0.0	 0.0	 1	 -30.0	 30.0;
 ];
 
+mpc.dcline = [
+	1	 10	 1	 10	 10	 25.91	 -4.16	 1.1	 0.92617	 10	 900 -900 900 -900 900	 0	 0	 0	 0	 0	 0	 0	 0
+]
```
ccoffrin: This fix belongs in PowerModels master.

REQUIRE
```diff
@@ -1,6 +1,8 @@
...
 InfrastructureModels 0.0.6 0.2-
 Compat 0.26.0
 Memento 0.6
+MathOptInterface
```
ccoffrin: This should specify a min bound.

src/io/matpower.jl
jd-lara: I did a cherrypick to update these lines to the current version in master and implement the fix for PWL costs in DC-lines. How would you prefer to update these files in the `moi-julia-v0.7`?

Issue #87, opened 2017-03-10
Computation of Power Transfer Distribution Factors for DC Models


ccoffrin:
Add a function for this.

ccoffrin:
Should be covered now by available features.

PR #365, opened 2018-08-20
Basic Network Flow and Copper Plate Models

### LINKED ISSUE #359: Copper Plate and NetworkFlow Approximations
Add these.

ccoffrin:
@claytonpbarrows, @kdheepak, @jd-lara

This adds the traditional active power only approximations called network flow / transportation and copper plate.

Network flow is a symmetric active-power-only formulation called `NFAPowerModel` and is tested for OPF problems.

After thinking about it a bit, I concluded the copper plate is *not* a power flow formulation.  It is simply a constraint called `constraint_power_balance`, which ensures the power injected equals the load in the system, including shunts.  Note this constant's implementation is currently active-power-only but it can be generalized to reactive as well.  To test this constraint, I added a new _Optimal Power Balance_ (OPB) problem, which finds the optimal generator dispatch to meet the loads, ignoring the network entirely (i.e. a copper plate model).

Both models are tested with PWL cost functions and the multi-network feature looking to possible time series models.

Although these formulations are active-power-only approximations, it is also worth noting that I have plans to implement the [relaxation variants](https://arxiv.org/abs/1506.05202) of these methods soon-ish.

Will keep this PR open for 24h for discussion.

Closes #359


frederikgeth:
Cool, great addition, this is very valuable for debugging. (By the way, why 'SymmetricDCPForm' and not 'DCPlosslessForm'? )

ccoffrin:
I don't have a strong preference for `SymmetricDCPForm` or `DCPlosslessForm`.

Maybe the type tree should start with active power only, then lossy and losses?

jd-lara:
I would sugges to keep `losless` since it is more descriptive for Power System's users.

ccoffrin:
The network flow formulation has explicit variables for line power flows but no DC power flow constraint (i.e. we omit the constraint `p_{i,j} = -b_{i,j}(va_i - va_j)`).

I have not yet implemented PTDF, for any model.

jd-lara:
@ccoffrin without the PTDF how is the NFA calculating the flows on the lines?

ccoffrin:
@jd-lara, I think we are not on the same page with the semantics of the different formulations.  For me there is,

1. **DC Power Flow**: an active power only linear power flow model with phase angle and generator dispatch variables.  Line flow values are determined by a linear function of the phase angle variables.
2. **PDTF Power Flow**: an active power only linear power flow model with generator dispatch variables.  Line flow values are determined by linear combinations of generator dispatch variables.
3. **Network Flow / Transportation Model**: an active power only linear power flow model with generator dispatch and line flow variables.  In this case, Ohm's law is not enforced and power can flow freely in the network, subject to individual the line flow limits.

PowerModels has (1) and this PR adds (3).  I see PDTF as a performance enhancement of the DC Power Flow Model, so I don't see it as a high priority, until we start having performance issues with that model.

jd-lara:
@ccoffrin We agree mostly about the definitions. 

The only difference is that in our case network flow is the same as PTDF because in many cases is impractical to have a "free-flow" transportation model where the flows don't resemble in some way the physics of the interconnected AC system.

The PTDF formulation is used in several applications as network approximation in DC where is not necessary to calculate the bus angles. In UC and ED formulations is common practice to use this formulation. Furthermore, the PTDF formulation is fundamental in contingency analysis formulations.

ccoffrin:
Yup, I understand the value and uses of PTDF.  It has been a feature request for some time, see https://github.com/lanl-ansi/PowerModels.jl/issues/87.  It's just not the topic of this PR.  

If it is a high priority for you guys, I will implement it.  As a motivating test case, I need to see a practical example in PowerModels where the current DCPPowerFlow model is too slow or space hungry with a commercial LP solver (e.g. guorbi / cplex).  Then I can measure the scalability improvements for various mathematical variants of that model.

For context, note that PDTF has never been a high priority for me becouse it does not generalize to any nonlinear model, which is where all of my R&D takes place.  Even DC with line losses is out of scope for PTDF, as far as I know.

Issue #369, opened 2018-08-20
Conic SOC form


kersulis:
It would be nice to have a conic SOC relaxation form that solvers like Mosek can work with. I have an implementation that works for OPF, but it currently duplicates a bunch of the code in wr.jl and needs more testing. A couple questions while I'm putting together the PR:

- Suppose I call the new form `SOCConicForm`. I need the conic form of the objective, but would like to use almost all of the code in `wr.jl` that currently dispatches on `T <: AbstractWRForm`.  So I basically want both `SOCConicForm <: AbstractWRForm` and `SOCConicForm <: AbstractConicPowerFormulation`. Would I address this by moving a bunch of `wr.jl` into `shared.jl` and dispatching on a type union?

- Does the new form belong in wr.jl?

ccoffrin:
I would put the new `SOCWRConicForm` in `wr.jl`.  It should be a subtype of `AbstractConicPowerFormulation` and be part of the `AbstractWRForms` type union.  This should get you basically everything needed for the default OPF problem.  You then need to implement `variable_voltage` and `constraint_voltage`, the first can be a copy of the `AbstractWRForm` the later you need to translate to conic form.

PR #371, opened 2018-08-20
Conic form of SOC relaxation

kersulis:
Addresses #369. With this form it's possible to solve an SOC OPF relaxation with a conic solver like SCS or Mosek.

I had to increase the maximum number of iterations for SCS to get a couple test cases to converge. I also had to increase objective tolerance for a couple cases, particularly the 5-bus case with negative generators. The PWL costs test case fails for some reason.

### DIFF
--- a/src/form/shared.jl
+++ b/src/form/shared.jl
@@ -18,7 +18,7 @@
 # instead.
 #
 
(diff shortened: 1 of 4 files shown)


ccoffrin:
Implementation looks reasonable to me.

PR #372, opened 2018-08-20
Jd/moi julia v0.7

jd-lara:
Update package manager files for Julia v0.7. Removes the old REQUIRE and adds the new TOML files. Also comments out the testing for now making the minimal test loading PowerModels.


ccoffrin:
If you think this is at a resonable point, I will merge it.

The travis run looks resonable, not sure why it returns 1 instead of 0.

jd-lara:
Thia is enough to make it work. The tests fail because all the scripts are commented out and nothing returns a true.

PR #380, opened 2018-09-04
Minor Tidying of WRM Forms

ccoffrin:
Renamed the SDP Decomp formulation Sparse because this is consistent with Julia naming conventions.  Reorganized the order of `wrm.jl` to be similar to other formulation files.

@kersulis look reasonable to you?


kersulis:
Looks great!

Issue #381, opened 2018-09-04
SparseSDP Documentation


ccoffrin:
Similar to `SDPWRMPowerModel`, `SparseSDPWRMPowerModel` would benefit from some documentation.

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/wrm.jl#L148

@kersulis, I think you are best suited to select the best reference.

BabakT1:
Hello, 
How can we get bus voltage values from the solution of SparseSDP?
I'm using "result=run_opf(network_data, SparseSDPWRMPowerModel, Mosek.Optimizer)
                  print_summary(result["solution"])"


Thanks,

ccoffrin:
In general there is no way to recover consistent "bus voltage values" from a convex relaxation (there is a special case however if the SDP matrix is Rank 1).  In this case you can use the `w` values are on the buses as an approximation of the voltage magnitude squared if you like and some algorithms have been proposed for recovering an approximate from a relaxed solution but none are sufficiently simple to be explained in a few sentences.

For future reference, questions of this type are best discussed on Julia's discourse forum, https://discourse.julialang.org/c/domain/opt/13

ccoffrin:
This is a good point the API is a little inconsistent in this regard.

They will appear in the diagonal elements of the WR matrices, but the mapping back to the bus ids can be non trivial.

PR #385, opened 2018-09-06
Citations for sparsity-exploiting SDP OPF

kersulis:
Addresses #381 

@molzahn are these appropriate references?

### DIFF
--- a/src/form/wrm.jl
+++ b/src/form/wrm.jl
@@ -145,7 +145,39 @@ end
 
 abstract type SparseSDPWRMForm <: SDPWRMForm end
 
-""
+"""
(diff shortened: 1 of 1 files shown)


molzahn:
Yes, these are the original references, as far as I am aware. Additionally, my 2013 paper was the first to discuss the clique merging ideas in the context of the OPF problem, although some related ideas were previously proposed by Kojima and his colleagues as well for general SDP problems. I can find that paper if you want to include it in the clique merging section of the code.

ccoffrin:
Thanks Jonas and Dan.  As I understand it, the current base implementation is not doing clique merging.  So it would seem that these references are sufficient.

PR #395, opened 2018-10-12
Support for Matlab runs

rb004f:
Adjusted the matpower export to protect from NaN values and run in matpower


ccoffrin:
Instead of this pattern, `dict["key"] != NaN ? dict["key"] : 0` what do you think about adding a function like this?
```
function get_default(dict, key, default=0.0)
if haskey(dict, key) && dict[key] != NaN
return dict[key]
end
return default
end
```

rb004f:
Sure.  This will also cover the (haskey(bus,"lam_q") ? bus["lam_q"] : "") cases we have in there.

rb004f:
Yes, the issue with NaN's arose from Matpower, not matlab

rb004f:
@ccoffrin your suggestions have been applied

ccoffrin:
Looks good.  Thanks!

PR #396, opened 2018-10-12
Moi julia v0.7

jd-lara:
Removes more deprecations for Julia v0.7 that fill the logs and updates some of the code to MOI from the MOI Branch. Most of the `Matpower` tests are passing but still WIP.


ccoffrin:
FYI, soon the PowerModels master will move to Julia v0.7.  Then this branch will be only about the move to MOI.

jd-lara:
@ccoffrin Will you use any of what we did in this branch for the update in master?

ccoffrin:
To best answer that I need to understand more about how Julia packages work in v0.7.  I will surely use this branch as a template for the v0.7 update.

Issue #415, opened 2018-11-14
Cost function definition


lthurner:
I am trying to generate a valid powermodels data structure directly without going through a .m or .raw file for interfacing with pandapower (see e2nIEE/pandapower#207), but I am unsure about how to define the cost variable. The documentation says:

> When present, the gencost data is incorporated into the gen data, the column names remain the same.

But not each column in the gencost matrix has a name, as the cost function can have different lengths. I figured out that for polynomial costs the cost variable is defined as a list of the coefficients, e.g. 
> cost = [c2, c1, c0]

Does that mean a piecewise linear cost function is defined as:
> cost = [p0, f0, p1, f1]

?

Also: matpower supports reactive power costs, but from what I can see in the code, that is not supported yet by PowerModels, is that correct?

ccoffrin:
You can see here how PowerModels interperpates the Matpower cost model here,

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/io/matpower.jl#L317

I think the only truly required parameters are "model" and "cost"; but "ncost" might also be needed.  At the moment, I am not aware of any code that is using "startup" or "shutdown".

Also, as a general tool to review the dictionary that PowerModels builds, you can use the following commands,
```
network_data = PowerModels.parse_file("case.m")
display(network_data) # raw dictionary
PowerModels.print_summary(network_data) # a table-like summary
```

At the moment, PowerModels does not considered the reactive cost model in Matpower;  if you will use that feature, I will add it.  Lets just open a specific Issue for that feature request.'

lthurner:
Thanks! I was asking because I was getting an error when trying to define a linear function with

```
cost = [c1, c0]
```

I now noticed that the matpower parser converts the costs always to a quadratic function, so that a linear function would be modeled as:

```
cost = [0, c1, c0]
```
So I assume PowerModels always considers the objective function to be quadratic?

> At the moment, PowerModels does not considered the reactive cost model in Matpower; if you will use that feature, I will add it. Lets just open a specific Issue for that feature request.'

I don't currently need it myself, I am just trying to figure out what PowerModels can do so that I can properly set up the interface.

lthurner:
> If you run into any disadvantages of this more general function let me know and I can look into adding an objective that is more suitable for linear cost functions.

Thanks I will, but everything seems to work fine for now!

> I think the only truly required parameters are "model" and "cost"; but "ncost" might also be needed. 

[...]

PR #419, opened 2018-11-26
Taking NCOST into account when parsing matpower casefiles #415

lthurner:
As discussed in #415, I changed the matpower parser not iterate until the end of each row, but only for the number of parameters that are defined in NCOST. To test this, I added an additional column to two casefiles from the testsuites (case5.m for polynomial and case5_pwl.c for piecewise linear cost function). In the current implementation, this would leave to different results, because the new column would become part of the cost function. With the changes in the matpower parser, the additional column is ignored, as it should be if my understanding of the matpower casefile function is correct.


test/data/matpower/case5_pwlc.m
```diff
@@ -27,11 +27,11 @@
...
-	1	0.0	0.0	4	 -0.1	   0.0	  0.0	   0.0	  0.1	   0.0	  0.2	   0.0;  % tests zero cost
-	1	0.0	0.0	4	  0.0	 841.0	 20.0	 841.0	 100.0	 841.0	150.0	 841.0;  % tests constant cost
-	1	0.0	0.0	4	170.0	4772.0	231.0	6203.0	293.0	7855.0	355.0	9738.0;
-	1	0.0	0.0	4	 22.0	1122.0	 33.0	1417.0	 44.0	1742.0	 55.0	2075.0;
-	1	0.0	0.0	4	  7.0	 897.0	  8.0	1187.0	  9.0	1479.0	 10.0	1791.0;
+	1	0.0	0.0	4	 -0.1	   0.0	  0.0	   0.0	  0.1	   0.0	  0.2	   0.0	   2.0;  % tests zero cost
+	1	0.0	0.0	4	  0.0	 841.0	 20.0	 841.0	 100.0	 841.0	150.0	 841.0	   2.0;  % tests constant cost
+	1	0.0	0.0	4	170.0	4772.0	231.0	6203.0	293.0	7855.0	355.0	9738.0	   2.0;
+	1	0.0	0.0	4	 22.0	1122.0	 33.0	1417.0	 44.0	1742.0	 55.0	2075.0	   2.0;
+	1	0.0	0.0	4	  7.0	 897.0	  8.0	1187.0	  9.0	1479.0	 10.0	1791.0	   2.0;
```
ccoffrin: This is an interesting case becouse the x_5 value does have a matching y_5 value.  Is this an expectable Matpower file?  If so, it is fine as is.  If not, just add a y_5 value to make it valid.

ccoffrin:
Verified that this implementation matches the Matpower one, will merge.

test/data/matpower/case5_pwlc.m
ccoffrin: Matpower does not seem to mind this.

PR #426, opened 2018-12-01
Improved Support for Linear Cost Functions

ccoffrin:
@lthurner, once this is merged please double check that this fixes any issues you are having.

Also note, now that `standardize_cost_terms` is run every time data is read in, parsing JSON data should be more robust to fairly sloppy cost function data. 

CC @rb004f


lthurner:
Thanks a lot for adressing this so quickly Carlton! I will have a look at it asap.

PR #428, opened 2018-12-03
Reorganization of Power Formulation Type Definitions

ccoffrin:
Moving these into a common location so it is easier to understand the type hierarchy.

@jd-lara, in this is preparation to explore the refactoring you have suggested.


ccoffrin:
CC @frederikgeth, this may also be of interest to you.

frederikgeth:
this does make it easier to understand the type hierarchy indeed, good move.

PR #432, opened 2018-12-11
ADD: Support for Logging in Julia Standard Library

pseudocubic:
Adds support for the Logging library within Julia v0.7/v1.0's standard
library. Support for Memento will be maintained until Julia v0.6 support
is dropped, due to no logging available in the Base before Julia v0.7.
Removal of Memento has been made easier by separating Memento-
based tests into their own blocks, and by creating temporary macros
upon load of PowerModels in VERSION < v"0.7.0-".

All unit tests were converted, no unit tests were dropped in this
addition.

TODO:

- [ ] Package-level logging (separate logger for package)

### DIFF
--- a/src/PowerModels.jl
+++ b/src/PowerModels.jl
@@ -5,14 +5,13 @@ using InfrastructureModels
 using MathProgBase
 using JuMP
 using Compat
-using Memento
 
 using Compat.LinearAlgebra
 using Compat.SparseArrays
 
-if VERSION < v"0.7.0-"
-    import Compat: @__MODULE__
+import Compat: @__MODULE__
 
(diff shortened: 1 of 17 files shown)


kaarthiksundar:
@pseudocubic - using the native Julia v1.0 logging is it possible to set package specific logging levels? For instance, is it possible to set a logging level info for powermodels and some other logging level for another package that uses native logging in a program that imports both packages? I was not able to figure this out using the documentation.

pseudocubic:
I know that it is possible to have different contexts with e.g. `with_logger`, the question is still whether that can be applied to only a package in an easy way. I know also that it should be possible to filter LogRecord by package.

PR #435, opened 2018-12-12
FIX: Additional information in PTI parse errors

### LINKED ISSUE #434: PTI Parser Error Messages
The PTI parser error messages should be improved so that it is easier for a user to track down the source of the error. 

For example, if `parse(dtype, element)` (around line 273 in pti.jl) generates an error, the user should know what line of the file it occurred in and what field id generated the error.

pseudocubic:
Adds additional information to ErrorException thrown for
`parse_line_element!`; error now gives line number, field and value,
section, and expected type of data.

Adds warning for section changes that happen when there is additional
data in the line that are not comments. Unit test added for new warning.

Closes #434


ccoffrin:
Changes look good to me, make sure to add a note to the change log.

PR #437, opened 2018-12-14
Adding AbstractActivePowerFormulation Type

ccoffrin:
@jd-lara, @claytonpbarrows added the active power only type based on previous discussions.  

Please confirm you are happy with this it will get into the next release of PowerModels.


jd-lara:
The changes look good but they only address the addition of the `AbstractActivePowerFormulation`, this still doesn't add the `AbstractComplexPowerFormulation <: AbstractPowerFormulation` that will be the umbrella over what we generally call "AC Models"

ccoffrin:
Jose, please review the email discussion from earlier in the week.  That
proposal makes a case for why I was not planning to add the Complex type,
at least any time soon.

jd-lara:
@ccoffrin you are correct the previous comment was not aligned with the e-mail chain. 

The changes are with us, I understand the logic behind it, and even though I would prefer to have at AC/DC distinction, this approach unifies the network modeling semantics in all packages in a cleaner way. However, we can address this in PowerSimulations using TypeUnions.

PR #449, opened 2019-01-07
Fix/psse bus sourceid

### LINKED ISSUE #447: Bus Source Id in PSSE Data
The `source_id` for a PSSE bus should be a string of the bus id in PSSE.  The bus name is not part of it's unique id.

pseudocubic:
Corrects source_id for PSS(R)E PTI parsed buses to be a string of the
bus id.

Closes #447


ccoffrin:
LGTM

PR #458, opened 2019-01-27
Update developer.md

byronbest:
format error
hope I fixed it, preview was not useful


ccoffrin:
Looks reasonable to me.  Thanks!

Issue #169, opened 2017-08-27
Documentation of Multinetwork Features?


ccoffrin:
Or maybe these should be left for advanced users only at the moment.

ccoffrin:
Going to skip for now.

ccoffrin:
In fact, all of the documentation could use an update to reflect the latest API...

yeesian:
I'll get to it at some point; sorry I've been swamped by work at the moment.

ccoffrin:
No need to worry about this.  There is no rush.  Plus the multi network features may still evolve as we start to use them.

PR #461, opened 2019-01-29
Update documentation

yeesian:
As promised in #169


ccoffrin:
LGTM, any objections @kaarthiksundar, @pseudocubic?

kaarthiksundar:
This looks great. Thanks @yeesian. I have no issues with the PR

docs/src/network-data.md
```diff
@@ -11,10 +11,14 @@ The network data dictionary structure is roughly as follows:
...
-"version":"2",
 "baseMVA":<float>,
 "source_type":<string>,
-"source_version":<string>,
+"source_version":{
+    "major":<int>,
+    "minor":<int>,
+    "patch":<int>,
+    ...
+},
```
pseudocubic: This doesn't seem necessary, since we can just use a single string to represent a version. Is this to support some change I'm not aware of having been discussed?

pseudocubic: Ah, I wasn't aware that it did this. I will make a new Issue to fix it, and will revert the documentation when it gets fixed.

PR #465, opened 2019-02-10
ADD: PTI default values

ccoffrin:
Adds ability for parser to accept null values in PTI raw files in cases where the PSS(R)E PTI v33 manual specifies the default value for that field.

Removes warning message indicating default vmax/vmin values, because these are the default PTI values now, and missing NVHI/NVLO will be automatically populated.

Adds fields missing at end of lines, with value `""`, which will be replaced with defaults, if possible.

Fixes bug in parsing of transformers where any missing fields at the end of each of the lines would cause the parse to fail. Now those missing fields are accounted for and will be added as `""` values with option to be replaced with default values later in the parse.

Minimal changes to old unit tests (mostly changes number of parsed fields, which have increased in some cases with the above listed changes.

Updates changelog.

Adds unit tests for default values.


ccoffrin:
LGTM.

PR #468, opened 2019-02-22
Dispatchable Loads and Shunts

ccoffrin:
@frederikgeth @rb004f @pseudocubic @kaarthiksundar @tasseff

This is my first attempt to support the `dispatchable` parameter in loads and shunts, comments welcome.  Along the way I made a significant revision to how the KCL constraints work, no renamed `constraint_power_balance`.

Planning to integrate this in the next breaking release v0.10


ccoffrin:
May be of interest to @jd-lara as well.

rb004f:
If we are all in agreement here, we will follow a similar pattern in GasModels... this will be a natural way to define slacks.  Given the name change, please keep GasGridModels.jl in sync since it calls many of the powermodel constraint functions.

frederikgeth:
looks great!

It's probably best to update the documentation on the [bus model](https://github.com/lanl-ansi/PowerModels.jl/blob/master/docs/src/formulation-details.md) with these new features as it deviates from the canonical AC OPF problem definition. 

Some power people will be confused by the concept "dispatchable" loads, as they would rather call that "sheddable" (where the default is that loads are dispatched, and the decision is to shed them). I don't think this is an issue though, as long as this noted somewhere in the documentation as well.

ccoffrin:
Idea: see if there is an easy way for each component to add its contributions to the power balance expression.  This would allow extensions packages to add new components without the need to overload the KCL constraint.

ccoffrin:
closing due to stale branch.  Will keep the branch around for a re-implementation at some point.

PR #471, opened 2019-03-07
FIX: PSSE string splitting

pseudocubic:
Fixes three separate issues in the parsing of PSSE text fields:

1. commas present inside quotations
2. reserved comment character `/` inside quotations 
3. quotations not getting stripped in cases where there is spaces before or after the quotation mark

As a consequence, PowerModels will now not parse cases where there are an uneven number of single quotes in a file, throws an error in cases where this is discovered.


ccoffrin:
Changes look good to me.  Good to merge once CI passes.

PR #479, opened 2019-04-04
multiconductor support for acr solution_builder

sanderclaeys:
acr.jl overloads add_bus_voltage_setpoint, but only implements single conductor support. This leads to unexpected behaviour in TPPM when implementing an ACR formulation.

This PR extends add_bus_voltage_setpoint to support multiconductor networks, analogous to the add_setpoint method in solution.jl.

**A small additional fix**
The current implementation relies on atan to recover the angle:
va = atan(vi/vr)
This returns an incorrect angle when the phasor is in the second or third kwadrant (shifted by 180 degrees). '[atan2](https://en.wikipedia.org/wiki/Atan2)' would give the correct angle.

### DIFF
--- a/src/form/acr.jl
+++ b/src/form/acr.jl
@@ -42,6 +42,7 @@ end
 "reference bus angle constraint"
 function constraint_theta_ref(pm::GenericPowerModel{T}, n::Int, c::Int, i::Int) where T <: AbstractACRForm
     @constraint(pm.model, var(pm, n, c, :vi)[i] == 0)
+    @constraint(pm.model, var(pm, n, c, :vr)[i] >= 0)
 end
 
 
--- a/src/form/acr.jl
+++ b/src/form/acr.jl
@@ -152,7 +152,7 @@ function add_bus_voltage_setpoint(sol, pm::GenericPowerModel{T}) where T <: Abst
             try
                 vr = getvalue(var(pm, :vr, cnd=c)[idx])
                 vi = getvalue(var(pm, :vi, cnd=c)[idx])
-                
+
                 vm = sqrt(vr^2 + vi^2)
 
                 sol_item["vm"][c] = vm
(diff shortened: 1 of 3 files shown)


ccoffrin:
Both good points!  The current implementation looks resonable to me.  Can you add a test of this formulation here, https://github.com/lanl-ansi/PowerModels.jl/blob/master/test/multiconductor.jl#L102?  Then I think we can merge this.

sanderclaeys:
There is a small bug in the acr overloading of constraint_voltage as well, where vmin and vmax are not indexed on the conductor. I fixed this in the latest commit.

sanderclaeys:
Adding a unit test for this is less trivial than it seems:
- In the current implementation, it is hard coded that the reference bus is initialized at theta=0, by setting vi==0. This is not 100% correct, because this allows for vr to be negative, leading to theta=pi. However, the current initial point given to ipopt (vr=1, vi=0) will go to theta=0.
- We need no to set a reference to [pi/2, 3*pi/2] to notice this bug.
- Doing this causes infeasibility issues. Ipopt seems to be quite sensitive to the starting point in rectangular coordinates. I ran into this issue when working on the ACR formulation in TPPM, and got it fixed by setting a better starting point than (vr=1, vi=0) for all phases.

So setting up a proper unit test requires fixing these issues in PowerModels, or patching all these issues in the unit test directly, for the purpose of the unit test alone.

sanderclaeys:
For now I added some patches in the new unit test to set an angle reference different from zero (pi) and to make acr converge.

ccoffrin:
These are all good points. The goals of the PowerModels tests is to test the basic multi-conductor software engineering functionality.  It is not a requirement that those tests be realistic problem formulations.  Realistic formulations should be tested in ThreePhasePowerModels.

Regarding the existence of two voltage points at the refrence bus, let's fix this by adding a constraint to https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/acr.jl#L43 for `vr >= 0`.  This constraint will need to overridden in ThreePhasePowerModels, but that is fine.

No need to worry about testing the `vr=-1` starting point in PowerModels, so you can drop that from the tests and focus on the `vr=1` case.

sanderclaeys:
I removed the new test altogether, because without a reference in the range [pi/2, 3*pi/2] essentially the test is identical to the test above it.

ccoffrin:
Ok great.  Do you mind restoring the multi conductor test?  It caught a bug
in constraint voltage and we don’t want a regression on that in the future.

sanderclaeys:
Right, I forgot about that. I added it again.

ccoffrin:
Looks good.  Thanks for the contribution!

PR #482, opened 2019-04-11
FIX: MultiConductorValue Broadcasting in Julia > v0.7.0-

pseudocubic:
In some cases broadcasting will result in a scenario where the Broadcasted struct field args is of type Extruded, which causes the find_mcv/find_mcm functions to fail. This Extruded structure contains the necessary information in the field x, instead of being in the structure of a tuple. This PR adds new functions to handle cases where the Extruded struct is utilized, e.g. when the element type is Real.

Also added is `norm()` for MultiConductorVectors, only for Julia 0.6, to be removed after Julia < v1.0 support is dropped. This is for a use case in ThreePhasePowerModels (transformers PR)

Unit tests added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- Fixed multiconductor broadcasting in Julia > v0.7.0-
-- Added Base.norm for multiconductorvectors for Julia < v0.7.0-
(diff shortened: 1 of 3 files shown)


ccoffrin:
Looks fine to me.  Do we expect any more updates related to TPPMs PRs before tagging a version?

pseudocubic:
I believe this covers the remaining test failures on TPPMs transformers PR

PR #485, opened 2019-04-15
ADD: JSON parser

### LINKED ISSUE #462: 'source_version' json export
JSON.jl converts version strings into `{"source_version": {"major": int, "minor": int, "patch": int ... }}`, instead of leaving as a plain string, making the loading of a json file inconsistent with the internal data format.

Tasks:
- [ ] Matpower parser
- [ ] PTI parser
- [ ] Documentation

### LINKED ISSUE #481: JSON parser
Need to add a json parser to support data structures with `MultiConductorVector` and `MultiConductorMatrix`.

Related: #462

pseudocubic:
Adds a json parser that handles Multi Conductor networks.

Also changes `"source_version"` to String, instead of a VersionString, so that when exported to a json it will remain a string, instead of being converted to a Dict with fields "major", "minor", and "patch".

`parse_json` can handle json strings directly

Updates `check_storage_parameters` to support multi conductor networks

Updates `parse_file` to open iostream, instead of having each separate parser open iostreams (allows for `parse_json` to be able to parse json strings directly)

Adds unit tests, updates documentation.

Closes #462
Closes #481


ccoffrin:
For testing that two PowerModels data are the same did you try `InfrastructureModels.compare_dict(source_data, destination_data)`?  If you overload `isapprox` for multi-conductor values you should be able to works around very minor differences in floating point numbers.

pseudocubic:
I thought there was a function like that, but since julia 1+ handles dict comparison now I had forgotten what it was. I have added it for Julia <= v"0.7.0-" tests.

test/multiconductor.jl
```diff
@@ -41,6 +41,59 @@ end
...
+        if VERSION > v"0.7.0-"
+            @test mc_strg_data == mc_strg_json_string
+        else
+            @test InfrastructureModels.compare_dict(mc_strg_data, mc_strg_json_string)
+        end
+
+        for comp in ["bus", "gen", "load", "shunt", "branch", "storage"]
+            for data in [mc_json_string, mc_json_file, mc_strg_json_string]
+                for (k, v) in get(data, comp, Dict())
+                    if k in mcv_fields[comp]
```
ccoffrin: Is this test needed.  What is it capturing that `InfrastructureModels.compare_dict` is not?

PR #491, opened 2019-04-17
Moi 2

jd-lara:
Removes Juniper from the package TOML files.


ccoffrin:
Are the MINLP tests active?

jd-lara:
Yes, they are, Juniper can be added just for testing as with the other solvers. What I am doing here is removing it from the main Project.TOML.

ccoffrin:
Makes sense.  This LGTM.  @pseudocubic, merge if you have no objections.

jd-lara:
I added Juniper in the test target in the Project.TOML the manifest is currently directing to Juniper#master since there is no official 0.3- release

jd-lara:
The way that the .yaml file is set-up doesn't load PowerModels.jl testing target as a project, so it never captures the required branch of Juniper to execute the MINLP tests. However, if the `--project` keyword is added to the Julia session for testing the current set-up passes all the tests. 

image

pseudocubic:
@jd-lara sorry, this took a bit to figure out, I didn't want the travis tests to be failing.

It looks like juniper will have to stay in `[deps]` until a new version that we can use is tagged. Apparently packages in `[extras]` are not git-cloned like those in `[deps]`, and julia thinks that there's no available versions of Juniper. Once Juniper gets another release, I will update the `project.toml` to remove Juniper from the deps section.

I think I have it working now, so once it passes in travis I'll merge. Thanks for the PR!

jd-lara:
@pseudocubic I suggest making Travis run the tests as a project such that you can have the flexibility needed.

pseudocubic:
@jd-lara I did think about that, but the travis documentation doesn't make it clear how to do this, and this solution was ultimately simpler, since only one line needs to change in the future.

If you can find a fix for the `.travis.yml` I would definitely consider another PR to fix it. As of now though this seems like the most straightforward and painless solution.

jd-lara:
@pseudocubic I can add the fix to this PR, we have that Travis configuration in `PowerSimulations.jl`

pseudocubic:
@jd-lara ah, thanks for pointing to an example, I see how its done now, for some reason I was thinking it would be harder. Feel free to make that change and if everything is still passing we should be good.

jd-lara:
@pseudocubic It seems that instantiate doesn't work in the same way when the Manifest points to a branch see https://github.com/JuliaLang/Pkg.jl/issues/1165. Yet, the travis.yml allows to instantiate and run a package within the CI.

jd-lara:
@pseudocubic @ccoffrin The CI tests are passing, there is an extra line in the travis.yml file to fetch the master branch from Juniper as a temporary fix while waiting for the release of the moi compatible version. 

The issue with having Juniper as a dependency is that it conflicts on the JuMP version with PowerSimulations' version and its dependency ParameterJuMP.

pseudocubic:
I am not sure why the macos julia 1.1 tests are stalling on travis, they work fine locally for me. i will take some time to debug. I am satisfied that this PR works correctly otherwise though and will merge it into moi-2

PR #492, opened 2019-04-25
Drop Support for Julia v0.6/v0.7

ccoffrin:


.travis.yml
```diff
@@ -3,25 +3,27 @@ os:
...
  directories:
    - /home/travis/.julia
 addons:
   apt_packages:
     - gfortran
 before_install:
-  - julia -e '(VERSION >= v"0.7" && using Pkg); Pkg.rm("PowerModels"); Pkg.rm("PowerModels")'
-  - julia -e '(VERSION >= v"0.7" && using Pkg); Pkg.update()'
+  - julia -e 'using Pkg; Pkg.rm("PowerModels"); Pkg.rm("PowerModels")'
+  - julia -e 'using Pkg; Pkg.update()'
```
pseudocubic: I would remove the `cache`, `addons`, and `before_install` sections (the latter will help nightly pass also I think). Recent testing seems to show that they aren't necessary, particularly for when you add the `Project.toml`

ccoffrin: Yup.  That's the plan.

ccoffrin:
I was hoping to _try_ getting away with a non-breaking version number, similar to InfrastructureModels.  We will see what the register says.  Can you check the version bounds I have proposed in the project.toml.  I didn't have time to think if that is the optimal way of translating them.

Project.toml
```diff
@@ -0,0 +1,36 @@
...
+[compat]
+Cbc = ">= 0.4"
+InfrastructureModels = "< 0.2"
+Ipopt = ">= 0.4"
+JuMP = "< 0.19"
+Juniper = "< 0.4"
+Memento = ">= 0.8, < 0.11"
+Pavito = "^0.1"
+SCS = ">= 0.4"
+julia = "^1"
```
pseudocubic: Remember that packages only in `extras` will cause an error if they are in `compat` (for now, I believe the Registrator needs a new release before we can add those to `compat`, which will hopefully happen soon)

ccoffrin: Good point.  Let's wait and see if that happens in the next few days before releasing.

jd-lara: You can handle the version requirements in the Extras using the Manifest.TOML temporarily.

ccoffrin: I'm too lazy to maintain that at the moment.  If the registator does not get fixed soon, we can just drop these for now.

PR #493, opened 2019-04-25
Infrastructure Models version update

jd-lara:
This is needed to have compatibility between IM#moi-2 and PM#moi-2


jd-lara:
@claytonpbarrows this should solve the problem with IM in PowerSimulations.

pseudocubic:
good catch. I changed to use the caret syntax in compat, and when travis looks good i'll merge, later this evening I guess

PR #495, opened 2019-04-27
Fix Solve Method Scope

ccoffrin:


ccoffrin:
FYI @pseudocubic, this lead to an issue where `solve` would not be defined if a script had `using JuMP; using PowerModels`.  Julia was not very good about indicating that the function was getting overwritten.

PR #498, opened 2019-05-04
Fix ACR unbounded voltage bug

### LINKED ISSUE #497: Voltage magnitude bound in ACR formulation
This function should be modified to take a keyword argument `bounded=false`.
https://github.com/lanl-ansi/PowerModels.jl/blob/6e1584115680f54d46f8eda67db2ca6e94f28221/src/prob/pf.jl#L25

In the ACR implementation, this function implements the equivalent of the voltage magnitude bound in the ACP formulation. This should be disabled for the pf problem formulation.

ccoffrin:
closes #497.

@sanderclaeys, would you like to test this fix before I merge, or are you confident it will be fine?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- nothing
+- Fixed voltage bound persistence bug in acr formulation (#497)
 
 ### v0.9.7
 - Minor fix solve method scope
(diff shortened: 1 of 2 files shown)


sanderclaeys:
I will test it tomorrow with the loads extension in TPPM; it has a test case which used to fail without the current workaround

sanderclaeys:
Okay, all good. I will update the extension in TPPM once this is merged.

PR #503, opened 2019-05-07
Update PowerModels.jl

jd-lara:
addresses error when calling `with_optimizer` when JuMP and PowerModels are called in the same script


ccoffrin:
LGTM, thanks!

PR #505, opened 2019-05-09
Improvements to Ref

ccoffrin:
This is a refactor on how the `ref` data structure is built, with the goals of improved model build times and allowing problem formulations to extend the precomputed values before the post method is called.

Notable changes,
- The GenericPowerModel constructor now only builds a very basic version of `ref`, which does the fully automatic and generic transformation steps of `data` into `ref`.  See `build_generic_ref`
- `core_ref!(pm)` can be used to populate `ref` with the basic info that nearly all models will need. It is called after the GenericPowerModel construct and before the model `post` method in `run_generic_model`.
- problem specific `run` methods can optionally add functions to the list defined by `ref_extensions`, which will be executed after `core_ref!(pm)` and before `post`.

The problem formulations OTS, TNEP, and OPB provide example for how this feature can be used.  For example see, `ne_branch_ref!`.

On a slightly tangential point, also updated how the `dcline` variables are setup to remove the need for `arcs_dc_param` and further simplify `core_ref!`.

Comments, questions and suggestions are welcome.  This will be merged shortly after the MOI transition is made. 

CC @frederikgeth, @hakanergun @rb004f, @pseudocubic, @sanderclaeys, @kaarthiksundar

### DIFF
--- a/src/core/base.jl
+++ b/src/core/base.jl
@@ -339,8 +339,8 @@ function core_ref!(pm::GenericPowerModel)
     core_ref!(pm.ref[:nw])
 end
 
-function core_ref!(refs::Dict)
-    for (nw, ref) in refs
+function core_ref!(nw_refs::Dict)
+    for (nw, ref) in nw_refs
         # filter turned off stuff
         ref[:bus] = Dict(x for x in ref[:bus] if x.second["bus_type"] != 4)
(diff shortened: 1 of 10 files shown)


ccoffrin:
@pseudocubic any particular reason to close this?

Issue #361, opened 2018-08-17
Support for MOI Linear Models


ccoffrin:
Make sure the linear models are working in MOI.

Issue #280, opened 2018-05-10
Upstream Requirements for MOI Migration


ccoffrin:
Issues currently blocking further progress,
- [x] https://github.com/JuliaOpt/Ipopt.jl/issues/114
- [x] https://github.com/JuliaOpt/Ipopt.jl/issues/129
- [x] https://github.com/JuliaOpt/Cbc.jl/issues/53
- [ ] https://github.com/JuliaOpt/Pajarito.jl/issues/411
- [ ] https://github.com/JuliaOpt/Pavito.jl/issues/7
- [x] https://github.com/lanl-ansi/Juniper.jl/issues/100
- [x] https://github.com/JuliaOpt/SCS.jl/issues/109
- [x] https://github.com/JuliaOpt/SCS.jl/issues/113

chriscoey:
also https://github.com/JuliaOpt/Pavito.jl/issues/7

ccoffrin:
Migration status update.

https://discourse.julialang.org/t/jump-and-julia-0-7-1-0/12964

ccoffrin:
At some point it will also be important to discuss the following points,
- the use of the `with_optimizer` pattern and how it make require PowerModels users to explicitly import JuMP
- will it be possible to specify solvers at the `optimize!` call, or should they always be provided at JuMP model instantiation

ccoffrin:
These discussion points were addressed in a recent JuMP-Dev phone call.  `with_optimizer` can be exported by PowerModels to make it cleaner for users.  Solvers can be provided at the time of the `optimize!` call.

PR #507, opened 2019-05-13
Update to JuMP v0.19

ccoffrin:


ccoffrin:
passing on Linux (OS X builds are slow), closes #280, closes #361.

PR #513, opened 2019-05-15
ADD: originating psse component type to source_id

### LINKED ISSUE #511: PTI Shunt Source Ids
We should think about the best way to encode the shunt source ids so that it is clear if the shunt came from the "fixed" table or the "swiched" table.

pseudocubic:
Adds a string to the beginning of each `source_id` indicating the originating component type in the PSSE file.

Adds/updates unit tests.

Updates changelog

Closes #511


ccoffrin:
Just curious, what guideline are we using for determining what these strings should be?

pseudocubic:
I used the lowercase versions of the section headers already in the code:

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/io/pti.jl#L15-L18

ccoffrin:
Merged into the `v11-rc`, will get into master with that branch.

PR #514, opened 2019-05-17
FIX: Storage constraint warn

pseudocubic:
`LOGGER` was missing from `Memento.warn` in the storage constraints.

Adds Unit tests and updates changelog


ccoffrin:
LGTM

PR #515, opened 2019-05-17
FIX: Parse of Bool arrays and Inf/NaN values from JSON

pseudocubic:
ThreePhasePowerModels has instances where Bool arrays are used on transformers, adding support for Bool arrays in base JSON parser.

Adds support for outputting Inf/NaN values in JSON and parsing them back in (not natively supported by JSON) by outputting strings and using `parse` on string values whose eltype is not supposed to be String.

Updates changelog, adds unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ PowerModels.jl Change Log
 
 ### Staged
 - Refactor connected_components to accept arbitrary edge-types with `edges` kwarg
+- Add Bool to allowed data types in json parse of MultiConductorVector/Matrix
 
 ### v0.10.0
 - Update to JuMP v0.19/MathOptInterface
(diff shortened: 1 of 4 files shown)


ccoffrin:
Should we add a simple test?

pseudocubic:
I have added the ability to output and parse back in Inf/NaN values, in addition to the original bool update.

ccoffrin:
To double check my understanding, support for Inf and -Inf as strings will only work inside multi-conductor values, correct?  I think that is good becouse supporting it for a float that occurs anywhere could be difficult.  You might also consider supporting NaN as well as Inf.

pseudocubic:
This is correct, only these values are supported in MultiConductor structs. NaN is also supported in the current version, but its hard to easily test for it, since `NaN == NaN` is false.

ccoffrin:
Just use `isnan()` and `isinf().

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ PowerModels.jl Change Log
 ### Staged
 - Refactor connected_components to accept arbitrary edge-types with `edges` kwarg
 - Fixes bug in warnings inside storage constraints
+- Add Bool to allowed data types in json parse of MultiConductorVector/Matrix
+- Add handling of Inf/NaN values in JSON parser
```
ccoffrin: I think this second line should also qualify with "in MultiConductorVector/Matrix"

PR #519, opened 2019-05-20
Add Support for MOI Status

ccoffrin:
I recomend making these status parameters a standard across all of InfrastructureModels @pseudocubic, @kaarthiksundar, @rb004f, @tasseff


jd-lara:
I will also implement the same in PowerSimulations

ccoffrin:
Fantastic!

Issue #499, opened 2019-05-04
Consider Removing constraint_voltage


ccoffrin:
Along the lines of #498, possibly all voltage related constraints should be moved from `constraint_voltage` to `variable_voltage`.

PR #531, opened 2019-06-01
Update Voltage Constraint Name

ccoffrin:
@frederikgeth and @sanderclaeys, this update starts to work toward merging the SOCWR and SOCBF types and address #499.

I have renamed `constraint_voltage` to `constraint_model_voltage` and written some documentation about what its purpose is.  I also added `constraint_model_current` for the BFM formulations.  This constraint plays a similar role to`constraint_model_voltage` but in the BFM space.  For the moment `constraint_model_current` only includes the old `constraint_branch_current` constraint, but other formulations would include more constraints (e.g. the non-convex versions of the BFM that work on cyclic networks).

I also removed the errors of the form "the BFM model only works with ..." in favor of checking for the existence of variables in the `constraint_model_*` implementations.  I think the current error messages are consistent with other PowerModels messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
+- Update constraint_voltage to constraint_model_specific (#499)
 - Update Matpower "bus_name" to map to bus parameter "name"
-- Updated slope and intercept data structure (#521)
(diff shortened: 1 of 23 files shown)


ccoffrin:
I am going to merge this to get on with some other updates, but still open for discussion.

frederikgeth:
This looks good!

PR #548, opened 2019-06-21
Add Check Status to Data Loading Checks

ccoffrin:


ccoffrin:
@spmacdonald, I believe this will address your issue.

PR #552, opened 2019-07-01
Add equations for alternative storage modeling

noahrhodes:
Added new equations that all storage to be modeled as a mixed-integer linear program.  These equations are alternative representations of the complementary constraint: `sc*sd == 0.0` 
This constraint can also be written as: `sc_on + sd_on == 1` where sc_on and sd_on are binary.  The power limits are then written as: `sc_on*charge_lb <= sc` `sc_on*charge_ub >= sc` etc.   This allows the variables `sc` and `sd` to function with all other existing storage equations in power models.

The new binary constraints' function names are not good : `constraint_storage_complementarity_mixed_int`. What is an appropriate naming for these functions?

Equations/variables for `storage_on_off` are also added, similar to the implementation of `gen_on_off`.

A new `storage_loss` function is added for DCPPowerModel to model a lossless storage device. This is primarily added to have a linear function. Any other interpretations that create a linear loss model are welcome.

Current tests for DC models that use storage fail because of the lossless storage equation.  Should the tests be updated or should this constraint have a new name, e.g. `constraint_storage_lossless`?

No new tests have been added. What types of tests should be added to test that check storage_on_off, etc. are working properly?


src/core/constraint.jl
```diff
@@ -180,6 +180,19 @@ function constraint_storage_complementarity(pm::GenericPowerModel, n::Int, i)
     JuMP.@constraint(pm.model, sc*sd == 0.0)
 end
 
+""
+function constraint_storage_complementarity_mixed_int(pm::GenericPowerModel, n::Int, i, charge_lb, charge_ub, discharge_lb, discharge_ub)
```
ccoffrin: Lets call this `constraint_storage_complementarity_mi` (for mixed integer) and we will call the current formulation `constraint_storage_complementarity_nl` (for nonlinear).

To change the name of `constraint_storage_complementarity` to `constraint_storage_complementarity_nl` without breaking downstream codes we will need to make the constraint template of `constraint_storage_complementarity` generate a deprecation warning and then call the new template `constraint_storage_complementarity_nl`.

You will want to add a new version of the `strg` opf model that has the `mi` variant and add some tests for that case.  You should be able to solve the DC-mi version with Cbc and the AC-mi version with Juniper.

src/core/constraint.jl
```diff
@@ -180,6 +180,19 @@ function constraint_storage_complementarity(pm::GenericPowerModel, n::Int, i)
     JuMP.@constraint(pm.model, sc*sd == 0.0)
 end
 
+""
+function constraint_storage_complementarity_mixed_int(pm::GenericPowerModel, n::Int, i, charge_lb, charge_ub, discharge_lb, discharge_ub)
+    sc = var(pm, n, :sc, i)
+    sd = var(pm, n, :sd, i)
+    sc_on = var(pm, n, :sc_on, i)
+    sd_on = var(pm, n, :sd_on, i)
```
ccoffrin: style nit, add a space between variables and constraints.  Delicate use of spacing helps make code easier to read and understand.

src/core/constraint.jl
```diff
@@ -189,3 +202,14 @@ function constraint_storage_loss(pm::GenericPowerModel, n::Int, i, bus, r, x, st
     sd = var(pm, n, :sd, i)
     JuMP.@NLconstraint(pm.model, ps + (sd - sc) == standby_loss + r*(ps^2 + qs^2)/vm^2)
 end
+
+function constraint_storage_on_off(pm::GenericPowerModel, n::Int, i, charge_lb, charge_ub, discharge_lb, discharge_ub)
```
ccoffrin: This constraint should also force `ps` and `qs` to zero as well.  There will need to be a specialized version for the  version for the `AbstractActivePowerFormulation` type as `qs` is not defined in those models.

src/core/constraint_template.jl
```diff
@@ -662,6 +662,17 @@ function constraint_storage_complementarity(pm::GenericPowerModel, i::Int; nw::I
...
 end
 
+""
+function constraint_storage_complementarity_mixed_int(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
+storage = ref(pm, nw, :storage, i)
+charge_lb = 0
+charge_ub = storage["charge_rating"]
+discharge_lb = 0
+discharge_ub = storage["discharge_rating"]
+constraint_storage_complementarity_mixed_int(pm, nw, i, charge_lb, charge_ub, discharge_lb, discharge_ub)
```
ccoffrin: Given that they are hard coded as zero, let's omit `charge_lb` and `discharge_lb`,  if we have a compelling use case later on the data model should be updated and then it will be easy to add this back in.

src/core/constraint_template.jl
```diff
@@ -697,6 +708,15 @@ function constraint_storage_state(pm::GenericPowerModel, i::Int, nw_1::Int, nw_2
...
 end
 
+""
+function constraint_storage_on_off(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw)
+    storage = ref(pm, nw, :storage, i)
+    charge_lb = 0
+    charge_ub = storage["charge_rating"]
+    discharge_lb = 0
+    discharge_ub = storage["discharge_rating"]
+    constraint_storage_on_off(pm, nw, i, charge_lb, charge_ub, discharge_lb, discharge_ub)
```
ccoffrin: same note as above.  This should also include the bounds for `ps` and `qs` (see note above).

src/form/dcp.jl
```diff
@@ -137,6 +137,13 @@ function constraint_voltage_angle_difference_ne(pm::GenericPowerModel{T}, n::Int
     JuMP.@constraint(pm.model, va_fr - va_to >= angmin*z + vad_min*(1-z))
 end
 
+# Should this only be for DCPLosslessForm or general DC forms?
```
ccoffrin: This should only apply to sub types of `DCPlosslessForm` and I think it is a good solution for that model.  We can do something better for lossy models where convex quadratic constraints are viable.

src/core/constraint_template.jl
ccoffrin: style nit, add a space between data extraction and the function call.

src/core/constraint_template.jl
ccoffrin: same style note as above.

src/core/constraint.jl
noahrhodes: If ps=0, sd=0, sc=0, and qs=0 then standby_loss also must equal 0. 

Looking at the storage energy constraint 
https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/constraint.jl#L173

```julia
JuMP.@constraint(pm.model, se_2 - se_1 == time_elapsed*(charge_eff*sc_2 - sd_2/discharge_eff))
```

I think it makes sense to set ps and qs to zero, but leave sc and sd active to represent the storage standby loss.

ccoffrin: Good point.  I think that's the correct semantics in the a non-linear model.  In the special case of the a lossless model we can also put `sc` and `sd` to zero, which will strengthen the formulation.

src/core/constraint.jl
```diff
@@ -170,22 +183,71 @@ function constraint_storage_state(pm::GenericPowerModel, n_1::Int, n_2::Int, i::
...
+function constraint_storage_on_off(pm::GenericPowerModel, n::Int, i, pmin, pmax, qmin, qmax, charge_ub, discharge_ub)
+    z_storage = var(pm, n, :z_storage, i)
+    ps = var(pm, n, pm.ccnd, :ps, i)
+    qs = var(pm, n, pm.ccnd, :qs, i)
+
+    JuMP.@constraint(pm.model, ps <= z_storage*pmax)
+    JuMP.@constraint(pm.model, ps >= z_storage*pmin)
+    JuMP.@constraint(pm.model, qs <= z_storage*qmax)
+    JuMP.@constraint(pm.model, qs >= z_storage*qmin)
+end
```
noahrhodes: constraint_storage_on_off has been updated with variations:

AbstractPowerModel: sets ps and qs to zero when off. sc, sd are free to represent loss
AbstractActivePowerModel: only sets ps because qs does not exist
DCPlossless: sets ps, sd, sc to zero to tighten formulation.

@ccoffrin what files should these be placed in?

ccoffrin: The AbstractPowerModel version is the most general so that goes in base/constraint.  The AbstractActivePowerModel goes in form/apo.jl (near the current storage constraints).  The lossless version can go in form/dcp.jl

noahrhodes:
@ccoffrin Updates have been made.

Any other issues you see?

src/core/variable.jl
```diff
@@ -522,6 +522,75 @@ function variable_storage_discharge(pm::GenericPowerModel; nw::Int=pm.cnw)
...
+    variable_active_storage(pm; kwargs...)
+    variable_reactive_storage(pm; kwargs...)
+    variable_storage_energy(pm; kwargs...)
+    variable_storage_charge(pm; kwargs...)
+    variable_storage_discharge(pm; kwargs...)
+    variable_storage_complementary_mi(pm; kwargs...)
+end
+
+""
+function variable_storage_complementary_mi(pm::GenericPowerModel; nw::Int=pm.cnw)
```
ccoffrin: This should be called, `variable_storage_complementary_indicator`

src/core/variable.jl
```diff
@@ -522,6 +522,75 @@ function variable_storage_discharge(pm::GenericPowerModel; nw::Int=pm.cnw)
...
+        [i in ids(pm, nw, :storage)], base_name="$(nw)_sc", Bin,
+        start = comp_start_value(ref(pm, nw, :storage, i), "sc_on_start", 0)
+    )
+    var(pm, nw)[:sd_on] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :storage)], base_name="$(nw)_sd", Bin,
+        start = comp_start_value(ref(pm, nw, :storage, i), "sd_on_start", 0)
+    )
+end
+
+function variable_storage_on_off(pm::GenericPowerModel; kwargs...)
```
ccoffrin: This should be `variable_storage_mi_on_off` to accommodate the NL variant where `variable_storage_complementary_indicator` is not required.

test/data/matpower/case5_uc.m
```diff
@@ -47,3 +47,13 @@
...
 ];
 
+% hours
+mpc.time_elapsed = 1.0
+
+%% storage data
+%   storage_bus ps qs energy  energy_rating charge_rating  discharge_rating  charge_efficiency  discharge_efficiency  thermal_rating  qmin  qmax  r  x  standby_loss  status
+mpc.storage = [
+	 3	 0.0	 0.0	 20.0	 100.0	 50.0	 70.0	 0.8	 0.9	 100.0	 -50.0	 70.0	 0.1	 0.0	 0.0	 1;
+	 10	 0.0	 0.0	 30.0	 100.0	 0.0	 0.0	 0.9	 0.8	 100.0	 110.0	 110.0	 0.1	 0.0	 0.0	 1;
```
ccoffrin: In optimization solvers the constraints `<value1> <= <variable> <= <value2>`, can cause numerical issues in when `value1 == value2`, as is the case with qmin/qmax here.  It is prefered to have some non-zero range for the variables.

test/opf-var.jl
```diff
@@ -173,6 +173,29 @@ end
         end
     end
 
+    @testset "test ac opf" begin
+        @testset "5-bus uc storage case" begin
+            # work around possible bug in Juniper strong branching
+            result = PowerModels._run_uc_strg_opf("../test/data/matpower/case5_uc.m", ACPPowerModel, JuMP.with_optimizer(Juniper.Optimizer, nl_solver=JuMP.with_optimizer(Ipopt.Optimizer, tol=1e-4, print_level=0), branch_strategy=:MostInfeasible, log_levels=[]))
```
ccoffrin: Can the default version of `juniper_solver` work in this case?  I think this variant was a work around for an old bug that has since been fixed.

test/opf-var.jl
```diff
@@ -207,6 +244,20 @@ end
         end
     end
 
+    @testset "test mi dc opf" begin
+    @testset "5-bus case" begin
```
ccoffrin: nit, indentation.

test/opf-var.jl
```diff
@@ -207,6 +244,20 @@ end
...
+        result = PowerModels._run_strg_mi_opf("../test/data/matpower/case5_strg.m", PowerModels.DCPPowerModel, juniper_solver)
```
ccoffrin: In this case can use use CBC instead of Juniper?

test/multinetwork.jl
```diff
@@ -270,16 +270,37 @@ TESTLOG = Memento.getlogger(PowerModels)
...
-            for (n, network) in result["solution"]["nw"]
-                @test isapprox(network["storage"]["1"]["ps"], -0.0447992; atol = 1e-3)
-                @test isapprox(network["storage"]["2"]["ps"], -0.0596441; atol = 1e-3)
-            end
+            @test isapprox(sum(network["storage"]["1"]["ps"] for (n, network) in result["solution"]["nw"]), -0.180000; atol=1e-3)
+            @test isapprox(sum(network["storage"]["2"]["ps"] for (n, network) in result["solution"]["nw"]), -0.240000; atol=1e-3)
+        end
+
+        @testset "test dc polar mi opf" begin
+            result = PowerModels._run_mn_strg_mi_opf(mn_data, PowerModels.DCPPowerModel, juniper_solver)
```
ccoffrin: The cbc solver should work here, correct?

src/prob/test.jl
```diff
@@ -91,6 +91,65 @@ function _post_uc_opf(pm::GenericPowerModel)
     end
 end
 
+"opf with unit commitment, tests constraint_current_limit"
+function _run_uc_strg_opf(file, model_constructor, solver; kwargs...)
```
ccoffrin: I am thinkinking that this model variant does not bring us too much utility in terms of test coverage.  I propose to remove it.  It seems like all of the functionalities of this model are covered by the other models, or did I miss something?

The general goal of these tests is to exercise all of the code's features with a small collection of problem variants.

src/core/variable.jl
```diff
@@ -522,6 +522,75 @@ function variable_storage_discharge(pm::GenericPowerModel; nw::Int=pm.cnw)
...
+        lower_bound = min(0, max(inj_lb[i], ref(pm, nw, :storage, i, "qmin", cnd))),
+        upper_bound = max(0, min(inj_ub[i], ref(pm, nw, :storage, i, "qmax", cnd))),
+        start = comp_start_value(ref(pm, nw, :storage, i), "qs_start", cnd)
+    )
+end
+
+function variable_storage_indicator(pm::GenericPowerModel; nw::Int=pm.cnw, relax=false)
+    if !relax
+        var(pm, nw)[:z_storage] = JuMP.@variable(pm.model,
+            [i in ids(pm, nw, :storage)], base_name="$(nw)-z_storage",
```
ccoffrin: `"$(nw)-z_storage"` should be `"$(nw)_z_storage"`

src/core/constraint_template.jl
```diff
@@ -657,11 +657,29 @@ function constraint_storage_current_limit(pm::GenericPowerModel, i::Int; nw::Int
...
 ""
+function constraint_storage_complementarity_nl(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
+    constraint_storage_complementarity_nl(pm, nw, i)
+end
+
+"deprecated: name change to constraint_storage_complementarity_nl( ... )"
 function constraint_storage_complementarity(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
-    constraint_storage_complementarity(pm, nw, i)
+    Memento.warn(_LOGGER, "call to depreciated function constraint_storage_complementarity use constraint_storage_complementarity_nl")
+    return constraint_storage_complementarity_nl(pm, nw, i)
```
ccoffrin: This `return` is not required.

test/opf-var.jl
noahrhodes: The objective in case5_strg is quadratic, and CBC/GLPK won't solve it

test/multinetwork.jl
noahrhodes: the objective is quadratic.  Should I create a new storage matpower without a quadratic objective?

src/prob/test.jl
noahrhodes: @ccoffrin 
This test is the regular uc_opf problem, with added storage.  Because storage is added, variable_storage_on_off(), variable_storage_indicator(), and constraint_storage_on_off() are tested.

Once storage is added to the general problems, this test would no longer be necessary.  

Should this test still be included?

test/opf-var.jl
noahrhodes: Correct, it now works.

I also removed the extra juniper solver from https://github.com/lanl-ansi/PowerModels.jl/blob/cd543a09e2b8de5c2d2aee2059a64a03b9263b07/test/opf-var.jl#L157

test/multinetwork.jl
noahrhodes: Making the objective linear makes this test return `NUMERICAL_ERROR`
https://github.com/lanl-ansi/PowerModels.jl/blob/cd543a09e2b8de5c2d2aee2059a64a03b9263b07/test/multinetwork.jl#L274

I will change `case5_strg.m` to have linear gen_cost, and make this test alter the gen_cost data to run properly.

src/prob/test.jl
noahrhodes: the strg was merged into uc_opf and uc_mc_opf problems.

noahrhodes:
@ccoffrin 
A few minor things fixed.

Anything else?

src/core/variable.jl
ccoffrin: @noahrhodes, not resolved?

noahrhodes:
@ccoffrin 
I updated the name change.  Was that the only change?

ccoffrin:
Yes, if this passes CI I will merge and tag a new release.

ccoffrin:
It's passing on linux so merging.

Issue #535, opened 2019-06-04
binary variables with bounds break with CPLEX


rb004f:
Cplex.jl errors out if it is provided binary variables with bounds. This causes the tnep implementation to break when using cplex. Possible options

1. Post this as an issue in Cplex.jl
2. Remove bounds on binary variables defined in PowerModels.jl.  This will also require some refactoring of various functions in relaxations_schemes to check if the variable is binary, i.e. _ub = JuMP.is_binary(z) ? 1.0 : JuMP.upper_bound(z).

ccoffrin:
For sure lets do 1.

For 2, this might be a JuMP bug, @mlubin what do you think?  The basic issue is that `upper_bound(z)` on a variable with `binary=true` and no explicit bounds leads to "Variable 0_1_branch_ne[2] does not have an upper bound." (e.g. see  https://github.com/lanl-ansi/PowerModels.jl/pull/536)

@kaarthiksundar, I recall something similar coming up in the Juniper update.  Any suggestions on how to proceed?

mlubin:
> The basic issue is that upper_bound(z) on a variable with binary=true and no explicit bounds leads to "Variable 0_1_branch_ne[2] does not have an upper bound."

This is by design. Constraints are independent in MOI. Creating a constraint `x \in {0, 1}` doesn't automatically create separate constraints for `x >= 0` and `x <= 1`. It's valid to create a binary variable and add bounds, however, so the solvers shouldn't crash in this case. It's definitely appropriate to file an issue with Cplex.jl.

ccoffrin:
Thanks for the clarification Miles, the JuMP semantics makes sense.  

In this context, we have generic non-convex constraint relaxations that reason about the minimum and maximum values of a given variable to build the relaxation's constraints.  So I am thinking we should add a helper function that supports a variety of combinations of bounded and unbounded variables and returns a valid range.

PR #562, opened 2019-07-16
Binary variable -> Integer variables with upper/lower bounds

noahrhodes:
Some solvers (juniper) don't like some constraints that can be used on binary variables, e.g.

`sum(z_branch) < 10`

because this implies a bound on the value of the binary variables. (this might be at some point in the magic of MathOptInterface, that a bridge attempts to bound a binary variable).  

Making these indicator variables integer with bounds of [0,1] prevents this problem.


ccoffrin:
This was recently changed.  See the discussion here, https://github.com/lanl-ansi/PowerModels.jl/issues/535

Sounds like a bug in Juniper?

ccoffrin:
In light of https://github.com/lanl-ansi/Juniper.jl/issues/143 can we close this issue?

If you are not actively working on another PowerModels PR, I would recomend re-forking the PowerModels repo to clean up these extra commits.

Issue #339, opened 2018-08-02
Add Switch Component


ccoffrin:
A breaker is a discrete operation that ties two bus voltages together.

If a line has `x=0` and `r=0` throw an error and suggest it should be modeled as a breaker.

Requested by @frederikgeth.

ccoffrin:
Features
- Status value 0/1 (conductorless)
- State value 0/1 by conductor (encoding open/closed resp.)
- Current limit (optional)
- MVA Limit (optional)
- Initial state
- Dispatchable
- Detect switches in datasets based on small r/x values

ccoffrin:
Use dispatchable for fixed vs non-fixed switch components.

ccoffrin:
closed by #570

PR #570, opened 2019-08-10
Switches

ccoffrin:
@frederikgeth, @rb004f, @pseudocubic a proposal for adding switch components to the data model.  If there are not major objections I'll add a few more formulations and then merge.

Core features of the component features are discussed here, https://github.com/lanl-ansi/PowerModels.jl/issues/339.


pseudocubic:
LGTM +1

ccoffrin:
Consider adding a node-breaker OTS example model and formulation.

frederikgeth:
The thermal limit property suggest P^2 + Q^2 <= (thermal rating)^2, but this isn't part of the switch model. I don't feel this constraint is needed though, and box constraints should suffice for almost all use cases. 

LGTM+1

ccoffrin:
@frederikgeth, you make a good point that the limit is not explicit in formulations and it should be.  

I agree that box constraints would be sufficient for our purpose but I failed to come up with a physical interpretation for them, thermal/current on the other hand has a well defined physical semantics that is intuitive, for me at least.  thermal/current limits also have the advantage that the user only needs to provide 1 number, rather than 4 in box bounds case.

Another option is to remove the limits from the Matpower model and use some pre-processing to compute reasonable values internally.  But I think it's better to ask the user to provide a resonable limit, even if it is quite large.  As I understand it there will always be a protection setting or a desired max loading that can motivate the need for such a limit.

ccoffrin:
@frederikgeth, I am going to merge this so that I can integrate the updates into the v13-rc, but I would still like to come to a consensus on the finer points of the model before tagging the next v12 release.

ccoffrin:
Closes #339.

Issue #568, opened 2019-08-09
Split Branch into Line and Transformer


ccoffrin:

ccoffrin:
This will also simplify the implementation of OLTC.

ccoffrin:
Revisit `tr-tp-models` branch once this is done.

ccoffrin:
Consider addressing #575 here as well.

ccoffrin:
This will also be a good time to revisit the naming of constraints like, 
```
constraint_ohms_yt_from(pm, i)
constraint_voltage_angle_difference(pm, i)
constraint_thermal_limit_from(pm, i)
```
Should probably have names like `constraint_power_flow_from`, `constraint_branch_thermal_limit_from`.

PR #574, opened 2019-08-14
matrix shunts

sanderclaeys:
This change will be needed to not break the existing unit tests in PMD once we move to matrix branch shunts.

### DIFF
--- a/src/core/data.jl
+++ b/src/core/data.jl
@@ -2322,15 +2322,15 @@ function make_multiconductor!(data::Dict{String,<:Any}, conductors::Int)
 end
 
 
-"feild names that should not be multi-conductor values"
+"field names that should not be multi-conductor values"
 const _conductorless = Set(["index", "bus_i", "bus_type", "status", "gen_status",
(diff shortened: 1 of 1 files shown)


pseudocubic:
@sanderclaeys be sure to check the math operations in multiconductor.jl and update any tests. I believe that some of the tests that involve MultiConductorVector math will need to be updated to use different fields because they were probably using `g_to` or the like from branches. We still want to test MultiConductorVectors though, so a different field should probably be chosen in those cases.

ccoffrin:
Agreed, other sections of the code and the tests most likely need to be updated.  This will be a breaking change so we will merge after the next v0.12 release.

sanderclaeys:
Maybe this should not even be part of PowerModels after all. It would also require for example rewriting constraint_voltage_angle_difference.

Instead, I will create a new make_multiconductor! method at PMD level, which adjusts the matrix shunts after the call to PM. This should be enough to fix the broken unit tests.

ccoffrin:
Sounds like a good plan to me.  We can revisit this in due course, especially during https://github.com/lanl-ansi/PowerModels.jl/issues/568.

PR #580, opened 2019-08-15
Storage Component Updates

ccoffrin:
@frederikgeth, working on the updates we have discussed.  Please review.

As I understand the most important point to update will be to add reactive standing losses.  What do you recomend calling these?  `p_standby_loss`, `q_standby_loss`?

CC @noahrhodes, FYI this may be of interest to you.

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -30,11 +30,12 @@ The list of columns for the generic storage model is roughly as follows,
   "qmax":<float, MVar>,
   "r":<float, p.u.>,
   "x":<float, p.u.>,
-  "standby_loss":<float, MW>,
+  "p_loss":<float, MW>,
+  "q_loss":<float, MVar>,
   "status":<int>,
 }
 ```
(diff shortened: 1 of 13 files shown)


frederikgeth:
I would rename standby to something more abstract, otherwise it is confusing to model EVs or thermal storage, or run-of-river pumped hydro. I was thinking of Pext for external? or Pother?

This parameter is a time series in case of multi-period forms. We'll need a second time series to indicate the inverter/converter status as well. E.g. the energy buffer can discharge while an EV is driving (Pext forces discharge), but it can't charge while driving (e.g. converter status = 0)

ccoffrin:
How about `p_loss`, `q_loss`, fairly generic and not prescriptive?

frederikgeth:
I can live with that name (although the 'loss' is allowed to go negative, e.g. when water naturally flows in into the upper reservoir of a pumped hydro plant, the energy stored increases, independently of the operation of the pump/turbine).

ccoffrin:
negative loss is fine by me as long as it has a reasonable motivation like you mention.  I'll work to that unless we think of something better in the next week or so.

ccoffrin:
Agreed this is nominally good.  Some revisions still may come before the final release.

PR #581, opened 2019-08-15
Optional Flow Limits

ccoffrin:
This feature is open for discussion for the next 48 hours or so.  The goal is to provide support to cases where only a subset of the branches have thermal limits.  In the new semantics existing problem definitions don't need to change; constraint templates like `constraint_thermal_limit_from` first check if a suitable limit value exists in the data model, if not, they skip adding that constraint.

In the case of things like `constraint_thermal_limit_from_on_off`, to be mathematically well defined, every branch must have a finite limit.  These constraint templates throw an error if a limit is not found.  The helper functions `calc_thermal_limits!` and `calc_current_limits!` can be used to add resonable limit values on all branches (these replicate the previous semantics).

Some benifits,
- Performance improvements are possible in models with very few binding limits
- It is possible to make a model with a mixture of thermal and current limits as specified in the data model

CC @jd-lara, @frederikgeth, @claytonpbarrows, @pseudocubic, @sanderclaeys, @rb004f, @hakanergun, @pseudocubic

### DIFF
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -451,7 +451,9 @@ function constraint_thermal_limit_from(pm::AbstractPowerModel, i::Int; nw::Int=p
     t_bus = branch["t_bus"]
     f_idx = (i, f_bus, t_bus)
 
-    constraint_thermal_limit_from(pm, nw, cnd, f_idx, branch["rate_a"][cnd])
+    if haskey(branch, "rate_a")
+        constraint_thermal_limit_from(pm, nw, cnd, f_idx, branch["rate_a"][cnd])
+    end
 end
 
 
@@ -466,7 +468,9 @@ function constraint_thermal_limit_to(pm::AbstractPowerModel, i::Int; nw::Int=pm.
     t_bus = branch["t_bus"]
     t_idx = (i, t_bus, f_bus)
 
-    constraint_thermal_limit_to(pm, nw, cnd, t_idx, branch["rate_a"][cnd])
+    if haskey(branch, "rate_a")
+        constraint_thermal_limit_to(pm, nw, cnd, t_idx, branch["rate_a"][cnd])
+    end
 end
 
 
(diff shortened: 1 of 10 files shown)


claytonpbarrows:
It looks like you are enabling this by checking for the existence of flow limit data... however this requires a user to add/remove that data to change the configuration of line enforcement. Would the capability be better enabled with an additional field such as "monitored"? Then users could add routines to set/unset the monitored status by voltage threshold or some other common metric.

ccoffrin:
@claytonpbarrows, thanks for your thoughts.  Your understanding is correct, the existence of the rating field is what is used to enable or disable the limit.

I am open to discussing using a different field such as `monitored = true/false` to control this.  However, my concerns for that approach would be,
- I think the term `monitored` has a slightly different semantics than no-flow constraint.  It is better not to mix meanings, even if the outcome is the same
- unless we add `monitored_mva` and `monitored_ma` it won't be possible to mix power and current ratings (maybe that is not too big if a deal)
- It will add additional bloat to the data model, which increases the amount of things a user needs to understand
- We will likely need to make a default semantics of `monitored = true`.  An advantage of this proposal is that there is no implicit semantics

jd-lara:
@ccoffrin If I got it right, these limits are physical limits related to the characteristics of the line. I think the proposed approach is correct since it specifies in the data model whether the modeler has interest in accounting for those.  It seems consistent with the PowerModels Data Model usage. 

On the other hand, I think that a monitored line might have a separate limit, typically in MW related to different aspects that tend to be system-wide. For instance, n-1 reliability constraints can require a line to operate the MW capacity with no relationship to the physical limit of the line. The type of constraint in such cases is a range constraint which would be superimposed over the quadratic constraints discussed here.

ccoffrin:
@jd-lara, you are correct and your explanation of monitored lines is also consistent with my understanding.  This is a feature worth consider but it is independent from this feature.

PR #607, opened 2019-09-21
added a relaxation option to variable_branch_indicator

rb004f:
A small change which adds a flag to variable_branch_indicator to allow for relaxed binary variables.  Follows the same pattern as the other indicator variables declarations


ccoffrin:
LGTM, will merge once tests pass.

PR #615, opened 2019-10-16
Added DCMPPowerModel Closes #612

### LINKED ISSUE #612: DC power flow unexpected results
I compared the pandapower and PowerModels.jl AC and DC power flow results for a small test grid (4 buses, 1 transformer, 2 lines). The AC results are equal. But when comparing DC results I noticed that the DC power flow voltage angle (Va) results differ. The difference is huge even for the small test case.

First I assumed it was a conversion error (from pandapower -> PowerModels.jl), but when forcing "br_r" values of all branches to zero, the results are equal:

```
    if model == "DCPPowerModel"
            for (i, branch) in pm["branch"]
                branch["br_r"] = 0.
            end
    end
    result = PowerModels.run_pf(pm, model, solver,
                                    setting = Dict("output" => Dict("branch_flows" => true)))
```

The matpower, pypower and pandapower implementations should not differ. Also I verified the formulas and "Wood, Allen - Power generation, operation and control" to be sure. Va should be independent of "br_r" since "br_r << br_x". However, on page 274 it is written that "Bik" is either "- 1 / br_x" or "- br_x / (br_r² + br_x²)" (I'm a little bit confused). Nevertheless, this leads me to the following two possibilities: 

1. The DC power flow results of PowerModels.jl are not correct, since "br_r" must not be considered in the DC power flow
2. The PowerModels.jl implementation just differs, since br_r is considered intentionally

I'm about to test other grids as well. I already noticed that the DC results of a (modified) version of the RTS (96 bus) grid still differ "a little" when forcing br_r = 0. I haven't figured out why yet. Maybe this is still a conversion error. 

P.S.: Thanks again @ccoffrin for the help via mail some weeks ago. I could fix a lot of conversion errors and also some bugs in the pandapower OPF :) The TNEP optimization works really well!

FlorianShepherd:
Added a DCMPPowerModel as discussed in #612  which takes into account:

- transformer TAP ratio
- transformer SHIFT
- neglects br_r (real part of branch impedance) 

in the DC calculation. This model can be used to replicate matpower (DC) results. 

Tests are added for TNEP and PF


ccoffrin:
@FlorianShepherd thanks for the PR!

The implementation looks good.  In terms of the tests, It would be preferable if we did not have to add a new case file, `matpower/case5.m` includes a 1 degree shift, can we use that as a test for the OPF formulation?  For TNEP test, I would recommend modifying `case3_tnep.m` or `case5_tnep.m` so that one of the branches has a phase shifter on it.  The other test values may need to be updated if a case file is changed, but this is ok as long as the results don't change too much.

FlorianShepherd:
I was afraid that you might write this. It's just some effort for me to convert the matpower *.m files to pandapower (we don't have a parser) and I do not use Matlab at all anymore to validate the results. But I understand that you don't want to add more test files.

ccoffrin:
If you are confident by this test that you developed that `DCMPPowerModel` is matching Pandapower (and by proxy Matpower), then you can simply switch to the PowerModels tests to use the data files that are already inside PowerModels and update the values to what comes up now.  The tests will ensure that we don't have regressions in the future.

FlorianShepherd:
Then I'll do that :) (I tested "my" four bus grid with tap, shift, tap+shift and without both. All results were equal and TNEP worked. I assume it works and change the test with one of the available grids.

FlorianShepherd:
done.

ccoffrin:
LGTM.  Will merge once CI passes.  Are you particularly keen to have a new release tagged or can we wait for a few more features to be added first?

FlorianShepherd:
Cool :) No I can wait and if needed use the github version. Thanks for asking!

PR #622, opened 2019-11-11
Current voltage form

frederikgeth:
This PR adds a power flow formulation in the current-voltage variable space (branches, mag+phase transformers, dclines supported).
- Defines formulation `AbstractIVRModel <: AbstractBFQPModel`, `IVRPowerModel <: AbstractIVRModel`
- Defines matching problem types `pf_iv` and`opf_iv`
- It is tested on a number of OPF and PF cases, up to `nesta_case2869_pegase`. IVR OPF seems to take about twice the calculation time as ACR, primarily due to a larger number of iterations. 
- Using it to solve a power flow sometimes results in ` LOCALLY_INFEASIBLE` even though it converges to the same solution as ACR/ACP (e.g. `case5_gap`). 

Looking for suggestions related to
- preferred function names and signatures for constraints/variables
-  ~~I substituted out the branch total current variables, and store them as expressions instead (series current is the fundamental variable). Adding an explicit total current variable is also possible (code in comments at the moment).~~ 
- Are you happy with the problem types?
- ~~Should we define a union type for ACR and IVR? Functions such as `constraint_voltage_magnitude_setpoint`, `constraint_theta_ref` are re-usable~~ 
- ~~The objective definition depends on nonlinear expressions, and you can't use `var(pm::AbstractPowerModel, nw::Int)` inside them. This breaks multiconductor support. What would a nice code pattern look like? (I will add the PWL variant following following a solution).~~ 
- ~~What is the cost of storing (a lot) of (NL)expressions? I could also store p/q variables as NLexpressions to simplify postprocessing.~~  

I look forward to your feedback to finish this PR.

~~I will afterwards need to update the implementation of this formulation when extending this in PMD to support the multi-conductor variant, as that is almost certainly unusable at the moment.~~ 

TODOs:
- [x] support PWL objective (and unit tests)
- [x] support multiconductor objective (and unit tests)
- [x] add docs for new problem types


frederikgeth:
Summary of recent changes:
- Complete PF and OPF support, multi-conductor-ready
- IVR as subtype of ACR
- PWL and NLP objectives supported
- no load current variable, i.e. substituted Pd and Qd into KCL constraint
- explicit total current variable at both ends of a branch
- rectangular power bounds imposed as part of the variable instantiation functions, i.e. for generators and dclines
- same approach for voltage bounds
- active and reactive power (branches, gens, dclines) stored as NLexpressions, to enable projection to power-polar voltage variable space, therefore no custom solution builder needed at this moment
- merged in changes from master

typical results, e.g. for `nesta_case2869_pegase.m`
```
ACP["objective"] = 133999.28806780474
ACR["objective"] = 133999.28808225528
IVR["objective"] = 133999.28808268235
ACP["solve_time"] = 13.339832053
ACR["solve_time"] = 9.636565385
IVR["solve_time"] = 42.880639609
ACP["termination_status"] = LOCALLY_SOLVED::TerminationStatusCode = 4
ACR["termination_status"] = LOCALLY_SOLVED::TerminationStatusCode = 4
IVR["termination_status"] = LOCALLY_SOLVED::TerminationStatusCode = 4
```
Ipopt iterations:
```
ACP: 35
ACR: 42
IVR: 88
```
The only case I encountered but couldn't get to match is `case5_gap` with an objective  of-27438.7, whereas ACP has -27497.7. Not sure what to make of this at the moment.

frederikgeth:
Ready for a round of review, whenever convenient @ccoffrin. Documenter does not manage to render the LaTeX in `math-model.md` correctly on my end. However, I can't even manage to render the version on master (which is deployed correctly)?! Not sure how to go about debugging it. My julia packages are seemingly up to date.

Also, if you have a script somewhere to run a comparison for all pglib cases, now is the time to point me to it :)

src/core/variable.jl
```diff
@@ -469,6 +446,66 @@ function variable_branch_current_rectangular(pm::AbstractPowerModel; nw::Int=pm.
...
+        p[(l,j,i)] = vrt*crt  + vit*cit
+        q[(l,j,i)] = vit*crt  - vrt*cit
+    end
+    var(pm, nw, cnd)[:p] = p
+    var(pm, nw, cnd)[:q] = q
+end
+
+
+"variable: `csr[l,i,j] + j csi[l,i,j] ` for `(l,i,j)` in `arcs_from`"
+function variable_branch_series_current_rectangular(pm::AbstractPowerModel; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true)
```
ccoffrin: It seems odd to me to include the `_rectangular` qualifier here.  Shouldn't the model type be encoding this?

src/form/iv.jl
```diff
@@ -1,126 +1,68 @@
...
-function constraint_voltage_magnitude(pm::AbstractIVRModel, n::Int, c::Int, i, vmin, vmax)
-    @assert vmin <= vmax
-    vr = var(pm, n, c, :vr, i)
-    vi = var(pm, n, c, :vi, i)
+"""
+Defines how current distributes over series and shunt impedances of a pi-model branch
+"""
+function constraint_current_to(pm::AbstractIVRModel, n::Int, c::Int, t_bus, f_idx, t_idx, g_sh_to, b_sh_to)
+    vrto = var(pm, n, c, :vr, t_bus)
+    vito = var(pm, n, c, :vi, t_bus)
```
ccoffrin: nit, `vr_to` and `vi_to` would be more consistent with the rest of the codebase.

src/form/iv.jl
```diff
@@ -1,126 +1,68 @@
...
-end
+    csrto =  -var(pm, n, c, :csr, f_idx)
+    csito =  -var(pm, n, c, :csi, f_idx)
 
-"reference bus angle constraint"
-function constraint_theta_ref(pm::AbstractIVRModel, n::Int, c::Int, i::Int)
-    JuMP.@constraint(pm.model, var(pm, n, c, :vi)[i] == 0)
-end
+    crto =  var(pm, n, c, :cr, t_idx)
+    cito =  var(pm, n, c, :ci, t_idx)
```
ccoffrin: nit, `cr_to` and `ci_to`.

src/form/iv.jl
```diff
@@ -152,109 +94,67 @@ function constraint_voltage_angle_difference(pm::AbstractIVRModel, n::Int, c::In
     vit = var(pm, n, c, :vi, t_bus)
```
ccoffrin: nit, `vr_fr`,`vr_to`,`vi_fr`,`vi_to` would be more consistent with the rest of the codebase.

src/form/iv.jl
```diff
@@ -152,109 +94,67 @@ function constraint_voltage_angle_difference(pm::AbstractIVRModel, n::Int, c::In
...
-    branch = ref(pm, n, :branch, f_idx[1])
-    g_sh = branch["g_fr"][c]
-    b_sh = branch["b_fr"][c]
-    tr, ti = calc_branch_t(branch)
-    tm = branch["tap"][c]
-
-    crf = JuMP.@NLexpression(pm.model, (tr[c]*csrfr - ti[c]*csifr + g_sh*vr - b_sh*vi)/tm^2)
-    cif = JuMP.@NLexpression(pm.model, (tr[c]*csifr + ti[c]*csrfr + g_sh*vi + b_sh*vr)/tm^2)
+    crf = var(pm, n, c, :cr, f_idx)
+    cif = var(pm, n, c, :ci, f_idx)
```
ccoffrin: nit, `cr_fr`, `ci_fr`

src/form/iv.jl
```diff
@@ -270,49 +170,40 @@ function constraint_current_limit(pm::AbstractIVRModel, n::Int, c::Int, f_idx, c
     crt =  var(pm, n, c, :cr, t_idx)
     cit =  var(pm, n, c, :ci, t_idx)
 
-    JuMP.@NLconstraint(pm.model, (crf^2 + cif^2) <= c_rating^2)
-    JuMP.@NLconstraint(pm.model, (crt^2 + cit^2) <= c_rating^2)
+    JuMP.@constraint(pm.model, (crf^2 + cif^2) <= c_rating^2)
+    JuMP.@constraint(pm.model, (crt^2 + cit^2) <= c_rating^2)
```
ccoffrin: nit, extra `()` in the constraint.

src/form/iv.jl
```diff
@@ -566,3 +378,48 @@ function _objective_min_fuel_and_flow_cost_polynomial_linquad(pm::AbstractIVRMod
...
+        pg_cost = var(pm, n)[:pg_cost] = JuMP.@variable(pm.model,
+            [i in ids(pm, n, :gen)], base_name="$(n)_pg_cost",
+        )
+        nc = length(conductor_ids(pm, n))
+
+        # gen pwl cost
+        for (i, gen) in nw_ref[:gen]
+            pg = [var(pm, n, c, :pg, i) for c in conductor_ids(pm, n)]
+            for line in gen_lines[i]
+                JuMP.@NLconstraint(pm.model, pg_cost[i] >= line.slope*sum(pg[c] for c in 1:nc) + line.intercept)
```
ccoffrin: Is `@NLconstraint` this the only difference here?  If so, why isn't `sum(pg[c] for c in 1:nc)` a quadratic expression?

src/form/iv.jl
```diff
@@ -566,3 +378,48 @@ function _objective_min_fuel_and_flow_cost_polynomial_linquad(pm::AbstractIVRMod
...
+        )
+        #to avoid function calls inside of @NLconstraint:
+        nc = length(conductor_ids(pm, n))
+        # dcline pwl cost
+        for (i, dcline) in nw_ref[:dcline]
+            arc = (i, dcline["f_bus"], dcline["t_bus"])
+            for line in dcline_lines[i]
+                #to avoid function calls inside of @NLconstraint:
+                p_dc = [var(pm, n, c, :p_dc, arc) for c in conductor_ids(pm, n)]
+                JuMP.@NLconstraint(pm.model, dc_p_cost[i] >= line.slope*sum(p_dc[c] for c in 1:nc)  + line.intercept)
```
ccoffrin: Same note from other objective.

src/prob/opf_iv.jl
```diff
@@ -1,53 +1,40 @@
 ""
 function run_opf_iv(file, model_constructor, optimizer; kwargs...)
-    return run_model(file, model_constructor, optimizer, post_opf_iv, solution_builder=solution_opf_iv!; kwargs...)
+    # return run_model(file, model_constructor, optimizer, post_opf_iv, solution_builder=solution_opf_iv!; kwargs...)
+    return run_model(file, model_constructor, optimizer, post_opf_iv; kwargs...)
```
ccoffrin: Default solution builder works becouse of expressions in `pm.var`?

ccoffrin:
I am also wondering if the introduction of the extra variables to model the loads (as per the original paper) resolves the runtime issue you found or makes no difference.  Do you have an intuition?

src/prob/opf_iv.jl
frederikgeth: indeed

src/form/iv.jl
frederikgeth: ok

src/form/iv.jl
frederikgeth: ok

src/form/iv.jl
frederikgeth: ok

src/form/iv.jl
frederikgeth: ok

src/form/iv.jl
frederikgeth: ok

src/core/variable.jl
frederikgeth: The functions in variable.jl do typically not dispatch based on the actual powermodel type, for re-usability, e.g.
```
function variable_branch_current_rectangular(pm::AbstractPowerModel; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true)
function variable_branch_series_current_rectangular(pm::AbstractPowerModel; nw::Int=pm.cnw, cnd::Int=pm.ccnd, bounded = true)
```
I can split them up into a real and imaginary part if you prefer that?

 In the current implementation, you have a from and to side total current variable, and a single (from-side) series current variable. Therefore, in iv.jl there is :
```
function variable_branch_current(pm::AbstractIVRModel; kwargs...)
    variable_branch_current_rectangular(pm; kwargs...)
    variable_branch_series_current_rectangular(pm; kwargs...)
end
```
 I can change that to : 
 ```
function variable_branch_current(pm::AbstractIVRModel; kwargs...)
    variable_branch_total_current(pm; kwargs...)
    variable_branch_series_current(pm; kwargs...)
end
```
but that would lead to potential confusion if you ever add `variable_branch_current_polar`?

File: src/form/iv.jl (hunk 7 above)
Comment (frederikgeth) [reply src/form/iv.jl]: :pg contains @NLexpressions, which can be nested into @NLconstraints only.

File: src/form/iv.jl (hunk 8 above)
Comment (frederikgeth) [reply src/form/iv.jl]: :p_dc contains @NLexpressions, which can be nested into @NLconstraints only.

**Comment by frederikgeth** [comment]
- having explicit load variables seemed to make little difference in calculation time (but I haven't tested rigorously)
- will take a look at the inf bounds and merge from master tomorrow.

File: src/core/variable.jl (hunk 1 above)
Comment (ccoffrin) [reply src/core/variable.jl]: Following the pattern from voltages in ACP and ACR.  A meta variable definition (e.g. `variable_voltage`) dispatches to different formulation agnostic variable definitions, e.g. `variable_voltage_angle `, `variable_voltage_magnitude ` / `variable_voltage_real`, `variable_voltage_imaginary`.  It is possible to follow a similar pattern here?

Comment (frederikgeth) [reply src/core/variable.jl]: Ok, I separated real and imaginary and put the construction of the active/reactive power expressions under `variable_branch_current`

**Comment by frederikgeth** [comment]
Ready for another review.

File: test/multiconductor.jl [hunk 10]
```diff
@@ -225,6 +225,21 @@ end
             end
         end
 
+
+        @testset "iv 5-bus case" begin
+            result = PowerModels._run_mc_opf_iv(mp_data, IVRPowerModel, ipopt_solver)
+
+            @test result["termination_status"] == LOCALLY_SOLVED
+            #TODO uncomment
```
ccoffrin: Should this part of the test be restored?

test/opf.jl
```diff
@@ -201,6 +201,56 @@ end
...
+        @test isapprox(result["objective"], 11625.3; atol = 1e0)
+        @test isapprox(result["solution"]["bus"]["1"]["va"], 0.0; atol = 1e-4)
+        @test isapprox(result["solution"]["bus"]["4"]["va"], 0.0; atol = 1e-4)
+    end
+    @testset "24-bus rts case" begin
+        result = run_opf_iv("../test/data/matpower/case24.m", IVRPowerModel, ipopt_solver)
+
+        @test result["termination_status"] == LOCALLY_SOLVED
+        @test isapprox(result["objective"], 79804; atol = 1e0)
+    end
```
ccoffrin: Would be good to add the new var bounds test here.  It would be,
```
    @testset "14-bus variable bounds" begin
        pm = build_model("../test/data/matpower/case14.m", IVRPowerModel, PowerModels.post_opf_iv)
        @test check_variable_bounds(pm.model)
    end
```

frederikgeth: Ok, this test fails as the IV form depends on `rate_a` being filled out. What should the behavior be?

test/multiconductor.jl
frederikgeth: partially restored, as it gives different results from ACP (not sure this test gives unique solutions?)

test/opf.jl
ccoffrin: The PowerModels data supports optional `rate_a` values now, so any code looking for this value should first check that it exists.  Here is an example of how I have handled it in the case of flow variable bounds, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/variable.jl#L326

test/multiconductor.jl
ccoffrin: As the IV formulation is non-convex, it is possible that Ipopt will converge to a different point than the other non-convex formulations.  When I implemented IV back in 2014 I found that it at had the same objective value in all cases as ACP/ACR.  So as long as that value is matching that seems to be consistent with past experiments.

frederikgeth: problem found and solved. Found some inconsistencies with how build_mc_data! builds the data model vs how PowerModelsDistribution expects it. Will open an issue

frederikgeth:
all tests now pass on my pc, ready to merge if travis passes

ccoffrin:
Thanks for the great contribution!

Issue #621, opened 2019-11-11
Wrong weights in objective variants


frederikgeth:
I think there are bugs here:
- https://github.com/lanl-ansi/PowerModels.jl/blob/3906996c383d7ce11c62791c611051b75edd76ae/src/core/objective.jl#L275
- https://github.com/lanl-ansi/PowerModels.jl/blob/3906996c383d7ce11c62791c611051b75edd76ae/src/core/objective.jl#L295
i.e. `cost_rev[2]` instead of  `cost_rev[3]` used for the quadratic term.

PR #623, opened 2019-11-12
Bugfix nlp objective weights

frederikgeth:
addresses https://github.com/lanl-ansi/PowerModels.jl/issues/621


ccoffrin:
In the future, if you post and issue and you plan to fix it as well, can you assign it to yourself?

frederikgeth:
Will do.

Issue #627, opened 2019-11-13
Proposal for more flexible result dictionaries


frederikgeth:
Formulations currently return different variable sets. E.g. SOC/SDP relaxations don't return voltage angles whereas ACP/ACR/ACT do. This was noted in https://github.com/lanl-ansi/PowerModels.jl/issues/6. 

Nevertheless, if the result of a SOC/SDP relaxation formulation satisfies its exactness guarantees, you can uniquely recover the voltage angles. On the other hand, if the relaxation is not exact, but you want to explore the result, it is currently hard to inspect the optimised values of *all* variables (e.g. wr and wi are not returned). 

Some observations
- ACP, ACR and ACT all return polar voltages
- To derive branch current, you actually want rectangular voltages
- To post-process from variable spaces P+jQ, W_i=U_i(U_i)* (e.g. includes common BFM and BIM forms) you need to know the reference bus and its reference angle (matpower assumes this is 0, but OpenDSS does not)

There is a use case for utilities to simplify conversion between variable spaces for formulations that are, or can be exact. I see 3 forms to standardize conversion utilities on
1. rectangular current and rectangular voltage
2. rectangular power and rectangular voltage
3. rectangular power and rectangular lifted voltage _+ reference bus angle(s)_
Alternatively, forms 1 and 2 can be defined with polar voltages (it adds a few redundant steps to sequential conversions, e.g. 3->1). 

Conversion between 1 and 2, and from 2 to 3 can be done node branch by branch/node by node in any order, i.e. it could be integrated in `solution_builder`. However from 3 to 2 you need to start from reference bus(es) and progress from there (order matters). This isn't easily integrated in `solution_builder`.

A mapping to existing formulations is
1. IVR
2. ACP, ACR, ACT, QC
3. SOCWR, SOCBF, SDPWRM, (QC?)

I see use cases for the following
- adding an option where each formulation returns _all_ of its variables
- by default a formulation returns the variables as defined in the above mapping (i.e. current situation)
- If you want to convert to another variable space, you need to call the right function yourself on a result dict. 

To do such conversion using only a result dict, you need a few things which currently are not stored in the result dict:
- from and to bus for each branch
- (list of) reference bus(es) and the(ir) angle(s)

I can spend a bit of time on this in the coming weeks, but I'm keen to get your thoughts before doing a lot of work.

ccoffrin:
[...]

PR #628, opened 2019-11-14
add info needed for variable space conversion to result, using setting

frederikgeth:
This is a possible first step in the direction of https://github.com/lanl-ansi/PowerModels.jl/issues/627. 

This PR adds a setting to have a self-contained result dict to enable conversion between variable spaces, using just the result dict. If enabled, it adds topology of all components, and angles of reference buses to result["data"].

- Do you think this is in scope?
- What would be a good name for the setting?
- It is off by default, is this preferred?

I'll add unit tests following feedback.

### DIFF
--- a/src/core/data.jl
+++ b/src/core/data.jl
@@ -2514,3 +2514,39 @@ function _make_multiconductor!(data::Dict{String,<:Any}, conductors::Real)
         end
     end
 end
+
+
+""
+function _get_topology(data)
+    topo = Dict()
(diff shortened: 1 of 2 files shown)


ccoffrin:
I would recomend holding off on work in this direction until I have a chance to respond to #627.  I have specific ideas on how I would like to see this.  I will endeavour to respond in the next 24 hours, 48 at the latest.

PR #636, opened 2019-12-14
Update SCS Settings to PMD values

ccoffrin:


ccoffrin:
Did not seem to work in this case.

Issue #637, opened 2019-12-15
Unused and untested functions in solution.jl


frederikgeth:
There seem to be a few bugs in `solution.jl`, where the old `add_setpoint` function name (wiithout bang) is used.

frederikgeth:
To summarize the problem, it seemed some functions were accidentally left behind in solution.jl, that are unused in PowerModels.jl, therefore the tests ran successfully. Removing them may nevertheless break modules that depend on PowerModels.jl. I took a quick look and I don't think there is a dependency in PMD.jl.

PR #638, opened 2019-12-15
Remove unused functions in solution.jl

frederikgeth:
Fixes https://github.com/lanl-ansi/PowerModels.jl/issues/637

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- Bugfix add_setpoint! (breaking) (#637) 
+- Removal of unused functions in `solution.jl`: `get_solution`, `add_generator_power_setpoint`, `add_storage_setpoint`, `add_branch_flow_setpoint`, `add_dcline_flow_setpoint` (breaking) (#637) 
 
 ### v0.13.1
 - Added DCMPPowerModel for replication of Matpower's DC model (#612)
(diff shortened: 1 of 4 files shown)


ccoffrin:
We need to add tests as well, to make sure the same bug does not resurface.

ccoffrin:
Also change log description should be slightly longer.

frederikgeth:
I added tests added for power flow result dicts (only OPF was tested).

The following functions are now removed as they were seemingly unused (that's why the tests didn't catch it) because they were superseded by new ones:
- get_solution -> solution_opf!
- add_generator_power_setpoint -> add_setpoint_generator_power!
- add_storage_setpoint -> add_setpoint_storage!
- add_branch_flow_setpoint -> add_setpoint_branch_flow!
- add_dcline_flow_setpoint -> add_setpoint_dcline_flow!
a merge error probably caused this bug? Note that this removal is potentially breaking.

ready for another inspection once Travis passes.

ccoffrin:
So it looks like the primary issue here was that there was some dead code that should have been removed in v0.12, but it was not.  Is that a fair assessment?  If so it's not a bug in the implementation.

frederikgeth:
Indeed. Other than updating the changelog and #637, is there anything more to address?

ccoffrin:
Nope, looks good.  I am just debating myself if we should treat this as a breaking bug fix (i.e. v0.13) or if we should put it in the next breaking release (i.e. v0.14).  I am leaning to including in v0.13.

frederikgeth:
Updated changelog. Ready to merge if travis passes

ccoffrin:
I will merge after getting v0.13.1 tagged.

PR #654, opened 2020-01-16
Bug Fix: check if stored energy variable exists

### LINKED ISSUE #653: Compent filtering causes issues for energy storage constraint
@ccoffrin 

When using a network where the status of a storage device might change, the variables for stored energy may not be created. 

Because `_ref_add_core!()` filters out inactive components, `ref()` will not contain any information about storage devices if `status=0`.

```julia
# Network 0 -> storage status is 0
_PMs.ref(pm, 0, :storage) 
  Dict{Int64,Any}()

# Network 1 -> storage status is 1
_PMs.ref(pm, 1, :storage)
   Dict{Int64,Any}(2 => Dict{String,Any}("energy_rating" => 1.0,"x" => 0.0,"energy" => 0.3,"r" => 0.1,"qs" => 0.0,"thermal_rating" => 1.0,"status" => 1,"damaged" => 0,"source_id" => Any["storage", 
2],"discharge_rating" => 0.7,"storage_bus" => 10,"charge_efficiency" => 0.9,"q_loss" => 0.0,"index" => 2,"ps" => 0.0,"qmax" => 0.7,"qmin" => -0.5,"charge_rating" => 0.5,"discharge_efficiency" => 0.8,"p_loss" => 0.0),1 => Dict{String,Any}("energy_rating" => 1.0,"x" => 0.0,"energy" => 0.2,"r" => 0.1,"qs" => 0.0,"thermal_rating" => 1.0,"status" => 1,"damaged" => 0,"source_id" => Any["storage", 1],"discharge_rating" => 0.7,"storage_bus" => 3,"charge_efficiency" => 0.8,"q_loss" => 0.0,"index" => 1,"ps" => 0.0,"qmax" => 0.7,"qmin" => -0.5,"charge_rating" 
=> 0.5,"discharge_efficiency" => 0.9,"p_loss" => 0.0))
```

This is a problem for the constraint  `constraint_storage_state(pm, i, n_1, n_2)` where the variable for the stored energy in the network _before_ the repair happens does not exist.  

Likely Fix: 
`constraint_storage_state(pm, i, n_1, n_2)` checks if  the stored energy variable for `n_1`, the previous time-step, exists and if not then call the function `constraint_storage_state(pm, i, nw=n_1)` for the initial storage state. 

I can make a pull request with this fix, if you agree that it is the correct approach.

noahrhodes:
In constraint_storage_state(pm,i,nw_1,nw_2) check
if ref(pm, nw_1, :storage, i) exists.  If the storage device had
status 0 in nw_1, this variable will not exist.  Instead, initialize
the stored energy from the network data.

Fixes #653 

@ccoffrin

### DIFF
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -849,7 +849,12 @@ function constraint_storage_state(pm::AbstractPowerModel, i::Int, nw_1::Int, nw_
         time_elapsed = 1.0
     end
 
-    constraint_storage_state(pm, nw_1, nw_2, i, storage["charge_efficiency"], storage["discharge_efficiency"], time_elapsed)
+    if haskey(ref(pm, nw_1, :storage), i)
+        constraint_storage_state(pm, nw_1, nw_2, i, storage["charge_efficiency"], storage["discharge_efficiency"], time_elapsed)
+    else 
+        # if the storage device has status=0 in nw_1, then the stored energy variable will not exist. Initialize storage from data model instead.
+        constraint_storage_state_initial(pm, nw_2, i, storage["energy"], storage["charge_efficiency"], storage["discharge_efficiency"], time_elapsed)
+    end
 end
 
 ""
--- a/test/multinetwork.jl
+++ b/test/multinetwork.jl
@@ -331,6 +331,9 @@ TESTLOG = Memento.getlogger(PowerModels)
             for (n,network) in mn_data["nw"]
                 delete!(network, "time_elapsed")
             end
+
+            mn_data["nw"]["1"]["storage"]["1"]["status"] = 0  # verify that storage activation does not cause error
+
             Memento.setlevel!(TESTLOG, "warn")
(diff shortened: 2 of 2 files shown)


noahrhodes:
One issue is what value should be used to set the storage value.

If storage is on , then becomes deactivated, then turned back on this will set storage value to what ever is in the network_data.  It will not link it to the last value of the stored energy.

Instead it could assume that the stored energy is 0 when the storage is activated, but then it is not possible to disconnect storage from the network and reconnect it with the same state of charge (or with standby losses). 

I am currently leaning towards setting the storage value to 0.  If just disconnecting from the network is desired, then adding a switching capability (or a branch with a switch) may be a better option.

ccoffrin:
In the long run I would like to support the case where the storage is disconnected from the system but the state of charge variables are still created so that we can model these sorts of things.  I agree formalizing the status vs state will likely resolve this.

In the meantime, I think this approach is reasonable to prevent crashes.  One request, I suggest that we throw a warning in the case when the storage device is not defined in `nw_1`, so that the user knows there may be an issue with the data.

ccoffrin:
@noahrhodes I am in a hurry to tag a new version, so I am going to merge this and add the warning.

noahrhodes:
@ccoffrin  thanks for adding the warning

PR #655, opened 2020-01-18
Fixing error with Cholesky for chordal extension with sparse SDP formulation

bcdandurand:


ccoffrin:
@bcdandurand thanks for this!  The changes look reasonable to me.  One addition that would be beneficial is a test that breaks in the old code and not this one, so that we don't have a feature regression if we refactor this section of code.  Do you know an easy way to build such a case?  For example via a specific kind of topology or component parameters?

bcdandurand:
I think the error with the old formulation should be encountered in constructing any power flow model with the IEEE Case 9 topology using sparse (chordal decomposition) SDP form. I can double-check this. The Case 9 topology happens to have a bipartite graph topology. Let me know if you have any questions. And thank you for incorporating this edit.

ccoffrin:
@blegat do you have comments on the pros/cons of using square-shaped vs triangular-shaped SDP constraints in this package?

bcdandurand:
In my local clone at: https://github.com/bcdandurand/PowerModels.jl
I'm currently using a modified version that uses triangular shaped PSD
constraints to replace the square ones in. I'm only doing this because I
also need to use the Dualization.jl package, which doesn't support
square-shaped PSD constraints; and also because I have not become deft yet
in the use of the MOI Bridge interface.

blegat:
As the matrix `[WR WI; -WI WR]` is not symmetric (since `WI != !WI`), you should use the square form.
By using the symmetric form, the `-WI` part is ignored so the constraint you pass is actually `[WR WI; WI WR]` instead of `[WR WI; -WI WR]`

ccoffrin:
Thanks @blegat!  @bcdandurand I suggest then we merge #657 and close this PR.  Sound good?

blegat:
Looking more closely at the changes, @bcdandurand added a constraint to ensure that WI is skew symmetric.
It might reduce the number of constraint and variable to create only the variables of the upper triangular part and then create a matrix with zeros on the diagonal, the variables created on the upper triangular part and their opposite on the lower triangular part. Then you have more than half the variables removed and no constraint created.

bcdandurand:
@blegat Yes, that is right. If you use the symmetric form, you need linear
constraints enforcing WI + WI^T = 0.  These constraints are implicitly
present in the square form by enforcing PSD. Once these constraints are
added,  then the symmetric (triangular) form may correctly be used. See:

https://github.com/bcdandurand/PowerModels.jl/commit/3bc7f75fcaa1c1d7e8b03a1c56a2b96f72a613b6


for example.

ccoffrin:
@bcdandurand, the basic fix is on master and slotted for the next release, so I propose we close this PR.  Feel free to open another one if you make other substantial improvements.

bcdandurand:
Thank you. Will do.

I reset my forked version of PowerModels to the upstream lanl-ansi master.
So it no longer contains the changes for using the triangular formulation
of the PSD constraints. If you decide to go with triangular form and you
would like me to add those changes back in and submit a pull request for
that, let me know.

PR #660, opened 2020-01-30
improve solution reporting for IVR

frederikgeth:
This changes the indexing of the series current variable for the IVR form, i.e. from `csr[(l,i,j)] for (l,i,j) in ref(pm, :arcs_from)` to `csr[l] for l in ids(pm, :branch)`.  Note that the series current is still defined in the i->j direction. This change makes it cleaner to report the variable to the solution builder.

This furthermore keeps the implementation consistent with IVR in PMD.


ccoffrin:
LGTM. However, will probably put this in v0.16.  Let me do the merge once the tests pass.

ccoffrin:
@frederikgeth, I don't see any other major breaking changes on the horizon so I am put this into the v0.15 release.

frederikgeth:
ok, great!

PR #659, opened 2020-01-29
V15 rc oltc

frederikgeth:
This extends v15-rc with tap magnitude and angle optimization, for just the ACPPowerModel

Keen to get your feedback.

This addresses https://github.com/lanl-ansi/PowerModels.jl/issues/543

Issue #543, opened 2019-06-15
Preliminary implementation of a controllable phase shifter


ccoffrin:
CC @jd-lara

ccoffrin:
@jd-lara, we added some support for this in v0.14.1, let me know if the current implementation suits your needs or not.

ccoffrin:
See the problem spec, `_run_opf_oltc_pst`.

PR #663, opened 2020-02-01
support for OLTC/PST

frederikgeth:
See discussion in https://github.com/lanl-ansi/PowerModels.jl/pull/659

Addresses #543


ccoffrin:
LGTM.  Thanks!

PR #666, opened 2020-02-06
Updates for PowerModels v0.15

### LINKED ISSUE #77: Re-work how solutions are reported
See if this can be simplified and if the the generic per-unit / mixed-units functions can be applied.

### LINKED ISSUE #405: Simplify MultiConductorValue Functions
For MultiConductorValues explore dropping overloaded functions like `+` in favor of dispatched versions `.+`.

### LINKED ISSUE #566: Allow add_setpoint to post a subset of variables
i.e. provide a list of damaged generators to `add_setpoint_gen_status!(...)` and only add this subset of generator variable values.

This can prevent attempting to index a variable that may not exist for all ids.

### LINKED ISSUE #575: Matrix Branch Shunts
Consider making branch shunts have the same matrix semantics as branch impedances.  E.g. see https://github.com/lanl-ansi/PowerModels.jl/pull/574.

### LINKED ISSUE #627: Proposal for more flexible result dictionaries

### LINKED ISSUE #648: _make_multiconductor! still has vector shunts instead of square matrices
_make_multiconductor! still has vector shunts, whereas PowerModelsDistribution switched it data model to matrix shunts. I reckon this should be brought into sync with the actual PMD datamodel?

### LINKED ISSUE #651: Standardize Setting Bounds in Variable Declarations
Initialize all variables as unbounded, and add bounds based on the status of the `bounded` flag.

ccoffrin:
closes #77, closes #405, closes #566, closes #575, closes #627, closes #648, closes #651


ccoffrin:
FYI @frederikgeth, @sanderclaeys, @pseudocubic, release is on its way.

jd-lara:
@claytonbarrows

ccoffrin:
@jd-lara, @claytonpbarrows, if you would like a hand updating your connector model I am happy to help.  At that level in the API, I _think_ the update will be fairly quick and painless.

frederikgeth:
solid effort, thanks!

ccoffrin:
#77 was a long time coming and well overdue.  Thank for the motivation to make it happen.

PR #676, opened 2020-02-24
Add linear branch flow model and branch flow storage problem type

peraaslid:
This PR adds the linearized branch flow model formulation for OPF and PF including tests. A storage problem for branch flow has also been added (`build_mn_opf_bf_strg`) and tested for both the new linearized branch flow and existing branch flow. `build_mn_opf_bf_strg` is the branch flow equivalent of bus injection model problem `build_mn_opf_strg`

### DIFF
--- a/src/form/wr.jl
+++ b/src/form/wr.jl
@@ -336,34 +336,6 @@ function variable_voltage_product_ne(pm::AbstractWRModel; nw::Int=pm.cnw, report
 end
 
 
-"do nothing by default but some formulations require this"
-function variable_current_storage(pm::AbstractWModels; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
-    ccms = var(pm, nw)[:ccms] = JuMP.@variable(pm.model,
-        [i in ids(pm, nw, :storage)], base_name="$(nw)_ccms",
-        start = comp_start_value(ref(pm, nw, :storage, i), "ccms_start")
-    )
-
-    if bounded
-        bus = ref(pm, nw, :bus)
-        for (i, storage) in ref(pm, nw, :storage)
-            ub = Inf
-            if haskey(storage, "thermal_rating")
-                sb = bus[storage["storage_bus"]]
-                ub = (storage["thermal_rating"]/sb["vmin"])^2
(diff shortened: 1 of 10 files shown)


src/form/wr.jl
```diff
@@ -337,7 +337,7 @@ end
 
 
 "do nothing by default but some formulations require this"
-function variable_current_storage(pm::AbstractWRModel; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
+function variable_current_storage(pm::AbstractWModels; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
```
ccoffrin: Same as above, `AbstractWConvexModels`.  Also now as a shared variable definition, it can move to `form/shared.jl`.

test/opf.jl
```diff
@@ -738,6 +738,70 @@ end
...
+        @test isapprox(result["objective"], 42565.8; atol = 1e0)
+    end
+    @testset "6-bus case" begin
+        result = run_opf_bf("../test/data/matpower/case6.m", LPBFPowerModel, ipopt_solver)
+
+        @test result["termination_status"] == LOCALLY_SOLVED
+        @test isapprox(result["objective"], 11277.9; atol = 1e0)
+    end
+    # @testset "24-bus rts case" begin
+    #     result = run_opf_bf("../test/data/matpower/case24.m", LPBFPowerModel, ipopt_solver)
```
ccoffrin: Curious, what is the issue with this case?

peraaslid: This case is infeasible for this formulations. I haven't figured out why.

peraaslid:
The intention with the linear branch flow approximation is to use it for operation and sizing of storages in LV/MV grids. I need a computationally efficient implementation for problems with many timesteps.

All your comments have been addressed and the code should be ready for a new review. Thanks for your feedback!

ccoffrin:
BTW, if you continue to work on your fork you will want to sync and start from the master as I squash merged this PR.

peraaslid:
Thanks! No need for tagging it right away, you can do it when you find it convenient.

Issue #684, opened 2020-03-14
Unify Branch Flow Expressions


ccoffrin:
See if PowerModels `expression_branch_flow_from/to` can be unified with PMSC's `expression_branch_power_from/to`  most likely the "has rating" check in `expression_branch_flow_from/to` should be moved into the problem formulation.

ccoffrin:
close in v0.16

PR #688, opened 2020-04-02
Remove Flow Variables from PF Definition

ccoffrin:
Also addresses #684.


jd-lara:
@claytonpbarrows this is breaking for us

ccoffrin:
@jd-lara, oh really? Are you using the Power Flow solver?

jd-lara:
We support it, but we try to get the flow variables. Not a bid deal.

PR #693, opened 2020-04-09
Fix some docstrings

raphaelsaavedra:
Small fix; there are a couple of useful functions that have incomplete docstrings (see https://lanl-ansi.github.io/PowerModels.jl/stable/parser/#PowerModels.parse_psse)


ccoffrin:
LGTM, thanks for the contribution!

More broadly the docs could use a lot of love in PowerModels.  This will not become a high priority for me until we reach v1.0.

PR #699, opened 2020-04-14
Fix reactive power control in storage systems

frederikgeth:
Due to a missing variable, the reactive power dispatch variables for storage systems were effectively fixed at 0. 

This PR adds the variable `sq`, which is bounded by the inverter rating, and fixes the tests with the new results.


ccoffrin:
When setting the variable bounds should we also check the "qmin" and "qmax" values in the data model?

frederikgeth:
Indeed, `qmin` and `qmax` can infer tighter bounds on `sq`. However I was afraid of breaking the extensibility in PMD doing that. I'm focusing on finishing the paper first, can leave this PR open or log an issue.

ccoffrin:
@frederikgeth, I am ready to go with this.  Will merge when CI passes.  I renamed the variable to `qsc` and I also move it to the other side of the equation, which makes the bound computations much simpler becouse the `qsc` bounds can be computed in an identical way to `qs`.  I updated the docs and technical report to reflect this change of sign.

frederikgeth:
looks great!

Issue #629, opened 2019-11-22
PSSE import - transformer impedance scaling and winding voltage


KSepetanc:
I tested PSSE network on polar (exact) OPF and after debugging and thorough PSSE documentation reading I found out it was transformer impedance problem. To get `Zeq` from PSSE data, `Z` from PSSE should be multiplied by `|t2|^2=("WINDV2"/"NOMV2")^2`. After doing so, power flow matched exactly PSSE calculations. I think PowerModels should automatically scale data into correct `Zeq`. The change is rather significant (10% easily in `R` and `X`).

I found about this in PSSE documentation: 
```
Program Application Guide - Volume 1 (v 33.5);
PAGV1.pdf; 
page 97 (section 4-7); 
equation 4.21b
available in PSSE installation folder
```

Additionally, having separate |t1| and |t2| is beneficial when dealing with tap position optimizations. Thus, I think `"WINDV1"`, `"WINDV2"`, `"NOMV1"` and `"NOMV2"` should be available when `import_all=true`.

So I propose changing tap calculation in `psse.jl` so it does not delete `"WINDV1"` and `"WINDV2"`:
line 387         `sub_data["tap"] = transformer["WINDV1"] / transformer["WINDV2"]`
and removing:  `"NOMV1"` and `"NOMV2"` from `_import_remaining!`.

ccoffrin:
@KSepetanc thanks for reporting this.  I don't have easy access to PSSE to verify solutions side-by-side, so I am not surprised that these kinds of inconsistencies exist.

I agree `import_all` should include all of the values, even the ones we translate into other fields.  So on that point a PR is welcome.

Concerning the more accurate calculation of `|t2|^2=("WINDV2"/"NOMV2")^2`.  I am a little hesitant to do this by default, becouse I am worried about the data quality across the wide variety of data sources that these networks come from. 

@pseudocubic, @frederikgeth any thoughts?

KSepetanc:
@ccoffrin I will try to explain briefly why scaling is needed. Impedance `Z` is something that should be available from nameplates which assume a bit more detailed replacement scheme not suitable for OPF (due to more variables). PSSE and PowerModels use the same simplified scheme for internal calculations (besides parallel shunt that is not relevant for this conversation). The simplified scheme is equivalent to the original only if `Z` is scaled into `Zeq` as described in the first comment.

[...]

PR #722, opened 2020-05-27
PSSE transformer Zeq scaling, winding and nominal voltage import all fix

KSepetanc:
Proposition to solve: PSSE import - transformer impedance scaling and winding voltage #629

I haven't done side by side tests for three winding transformers as my network didn't have any.
However, for two winding transformers I can confirm it was necessary to scale Z with |t2|^2 to get the same results as PSSE. This is described in PSSE manual cited in the issue thread.

Winding voltage and nominal voltage are required for manual implementation of tap changers and should be loaded when import all is specified.

### DIFF
--- a/src/io/psse.jl
+++ b/src/io/psse.jl
@@ -360,9 +360,23 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                     br_x *= (transformer["NOMV1"]^2 / _get_bus_value(transformer["I"], "base_kv", pm_data)^2) * (pm_data["baseMVA"] / transformer["SBASE1-2"])
                 end
 
-				# Tap Zeq scaling
-                sub_data["br_r"] = br_r * (transformer["WINDV2"]/transformer["NOMV2"])^2
-                sub_data["br_x"] = br_x * (transformer["WINDV2"]/transformer["NOMV2"])^2
+				# Zeq scaling for tap2 (see eq (4.21b) in PROGRAM APPLICATION GUIDE 1 in PSSE installation folder)
+                # Unit Transformations
+                if transformer["CW"] == 1  # "for off-nominal turns ratio in pu of winding bus base voltage"
+					br_r *= (transformer["WINDV2"])^2
+					br_x *= (transformer["WINDV2"])^2
+				else  # NOT "for off-nominal turns ratio in pu of winding bus base voltage"
+					if transformer["CW"] == 2  # "for winding voltage in kV"
+						br_r *= (transformer["WINDV2"]/_get_bus_value(transformer["J"], "base_kv", pm_data))^2
+						br_x *= (transformer["WINDV2"]/_get_bus_value(transformer["J"], "base_kv", pm_data))^2
+					else  # "for off-nominal turns ratio in pu of nominal winding voltage, NOMV1, NOMV2 and NOMV3."
+						br_r *= (transformer["WINDV2"]*(transformer["NOMV2"]/_get_bus_value(transformer["J"], "base_kv", pm_data)))^2
+						br_x *= (transformer["WINDV2"]*(transformer["NOMV2"]/_get_bus_value(transformer["J"], "base_kv", pm_data)))^2
+					end
+				end
+
+                sub_data["br_r"] = br_r
+                sub_data["br_x"] = br_x
 
                 sub_data["g_fr"] = pop!(transformer, "MAG1")
                 sub_data["b_fr"] = pop!(transformer, "MAG2")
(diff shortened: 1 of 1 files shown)


ccoffrin:
Thanks for the PR.  The changes do impact the unit tests, so this a significant change.  We should verify that the change improves consistency with respect to PSSE.  Can you do a side-by-side comparison on these two cases,
- "5-bus case", `data/pti/case5.raw` (unchanged control)
- "5-bus with asymmetric line charge case", `data/pti/case5_alc.raw` (changed breaking test)

If a power flow solve for these are consistent with PSSE, then the unit tests need to be updated to reflect the revised model.

KSepetanc:
I have temporary reverted to original PowerModels code regrading the Input all.

I have adapted the code to account for transformer["CW"] flag so it should work now.

I have done side by side tests on proposed networks:
modified secondary tap as the original had |t2|=1 and recomputed in PSSE;
fixed all voltage magnitudes and angles to the result of PSSE calculation and added on every bus virtual unlimited generator (P and Q as free variables) so the model is feasible (and removed P^2+Q^2<=S^2 equations);
all existing generators are also producing fixed quantities as calculated by PSSE;
virtual production is zero on forked PowerModels and nonzero around transformers with original.

This exercise confirms the fix for 2 winding transformers. I have also tested with another CW flag.
However, the networks do not contain 3 winding transformers so I can not confirm that case.

Regarding the import all, I would like to get tap1,2 and 3 which I basically calculated in the steps to fix Zeq scaling. Tap1,2 and 3 are generally not e.g. transformer["WINDV2"] as I thought initially.

How to export these calculated scaling values (tap1,2,3)?

Please note there is free PSSE Xplore which is basically normal PSSE for up-to 50 buses.

KSepetanc:
I tested side by side 3 winding transformers in PSSE and PowerModels on three_winding_test.raw.

Original PowerModels matches Z calculations in PSSE. This might be simplification on PSSE side (it seems that it doesn't do scalings as it does for two winding transformers).

However, I found another PowerModels bug in that test case. It does cause wrong power flows as a result:
Star bus is marked as inactive if primary winding is inactive;
This removes connecting active branches (secondary and tertiary) and thus the error;
Warning message occurs: 
[warn | PowerModels]: active branch 12 is connected to inactive bus 10004
[warn | PowerModels]: active branch 11 is connected to inactive bus 10004

Second error is that all star buses are always in the set of buses. This shouldn't happen when all windings are inactive.

I don't think I could tamper with bus notations for PR and it should be a separate PR.

Remained to be fixed as part of this PR:
Exporting data. I basically need |t2|. |Tap|=|t1|/|t2| and |t1| range is usually specified, i.e. 20 steps from 0.9 to 1.1 or similar. |t2| is needed to correctly implement OLTC.
Do you have suggestions how to deal with this problem?

Another question is how does PowerModels compute, and it does compute, star bus voltage magnitude and angle? (Vm and Va)

KSepetanc:
@ccoffrin can you help with this? Also, why did codecov fail? What does it do?

ccoffrin:
@KSepetanc, sorry for the delay!  This is not a blocking issue, but it is useful information.  This means that slightly less lines of code are covered by the tests than before.  This often happens when adding some lines of new code.  In this case it indicates that we should add a test network where these new winding computations are utilized.

Concerning the status issue, that should be addressed on the master branch.  You will want to sync your fork.

If you are happy with the constant calculations in this PR I am happy to merge it.  It seems like it does not impact the current tests, which is good.

As I understand it, the last remaining outstanding point is be able to get at the `WINDV` values when using `import_all=true`, I can take care of that on master for you.

Does that sound reasonable?

KSepetanc:
HI @ccoffrin . The last commit, a few moments ago just removes one empty line to make the code nicer.

Data needed to calculate |t2| of transformer which is needed for OLTC:
"CW"
"WINDV"
"NOMV"

It also needs the following data which is already available:
bus base_kv
source_id - "winding" (to know if it is 2 winding or 3 winding)

It is appreciated very much if you can fix this output data.

Summary regarding the star node bugs:
1) Star bus is marked as inactive if primary winding is inactive (should be all 3 inactive to be marked inactive);
2) All star buses are always in the set of buses regardless of activity status.

Star node bugs will change some test results. (at least three_winding_test.raw)

I an fine with pulling in this commit. I have checked side by side with PSSE.
2 winding calculations are now fixed. 3 winding calculations were correct in original master PM.

KSepetanc:
When CW=3 then NOMV# is needed as well since for that CW WINDV is defined in p.u. with base equal to NOMV. So, the |t#|=windv#*(nomv#/bus_base_kv).

Can you export that as well? Other things work and this is the last outstanding thing.

ccoffrin:
`nomv` on import all added to master.

KSepetanc:
It works!
Should be ready for a merge.
Thanks!

PR #723, opened 2020-06-03
Fix kwargs for `compute_ac_pf`

kdheepak:
- [x] Change `,` to `;` for `compute_ac_pf`.

Currently, `kwargs...` is splatted in the arguments, resulting in `compute_ac_pf(pf_data, args...;)` instead of the intended `compute_ac_pf(pf_data; kwargs...)`. 

This causes a `MethodError` if `compute_ac_pf(file; finite_differencing = true)` is called, because there's no method `compute_ac_pf(pf_data::T, ::Pair{Symbol, Bool})`.

I imagine similar changes might be needed elsewhere in the codebase.

### DIFF
--- a/src/prob/pf.jl
+++ b/src/prob/pf.jl
@@ -246,7 +246,7 @@ end
 
 function compute_ac_pf(file::String; kwargs...)
     data = parse_file(file)
-    return compute_ac_pf(data, kwargs...)
+    return compute_ac_pf(data; kwargs...)
 end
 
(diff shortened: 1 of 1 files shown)


ccoffrin:
Thanks for the report!  I push a more comprehensive fix to master with some tests as well.  Can you confirm the current master is wroking for you and then close this PR?

kdheepak:
It works! Thanks!

PR #724, opened 2020-06-13
LPAC TNEP

### LINKED ISSUE #708: LPAC in TNEP
Dear Carleton and colleagues,

I tried to use the LPACCPowerModel in a TNEP optimization and got:

```
ERROR: LoadError: MethodError: no method matching variable_ne_branch_voltage(::LPACCPowerModel)
Closest candidates are:
  variable_ne_branch_voltage(::AbstractACPModel; kwargs...) 
  variable_ne_branch_voltage(::AbstractDCPModel; kwargs...) 
  variable_ne_branch_voltage(::AbstractWRModel; kwargs...) 
```
as I've understood it, there is no definition of the required TNEP constraint for at least the branch voltage of new branches. I wonder why (is it not possible to define it or just not defined?)? Since I've found your paper: ` Transmission Network Expansion Planning: Bridging the gap between AC heuristics and DC approximations ` I thought it would be great to test the LPAC in a hybrid heuristic approach combining switching optimization and expansion measures. I already implemented it with the DC model to solve the TNEP, but, as you've also stated in the paper, the DC TNEP results are not solving the AC-problem most of the time. What I need are good initial solutions for the heuristic and your results look very promising. 

Maybe I can try to implement the LPAC for TNEP or do you have any other suggestions? 

Thank you in advance and keep up the great work. PowerModels.jl is cool :)!
Best,
Florian

jay-dave:
Fixes #708

### DIFF
--- a/src/form/lpac.jl
+++ b/src/form/lpac.jl
@@ -130,3 +130,249 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
     JuMP.@constraint(pm.model, q_to == -(b+b_to)*(1.0 + 2*phi_to) - (-b*tr+g*ti)/tm^2*(cs + phi_fr + phi_to) + (-g*tr-b*ti)/tm^2*-(va_fr-va_to) )
 end
 
+"`angmin*branch_ne[i] + vad_min*(1-branch_ne[i]) <= t[f_bus] - t[t_bus] <= angmax*branch_ne[i] + vad_max*(1-branch_ne[i])`"
+function constraint_ne_voltage_angle_difference(pm::AbstractLPACCModel, n::Int, f_idx, angmin, angmax, vad_min, vad_max)
+    i, f_bus, t_bus = f_idx
+
+    va_fr = var(pm, n, :va, f_bus)
+    va_to = var(pm, n, :va, t_bus)
+    z = var(pm, n, :branch_ne, i)
+
+    JuMP.@constraint(pm.model, va_fr - va_to <= angmax*z + vad_max*(1-z))
+    JuMP.@constraint(pm.model, va_fr - va_to >= angmin*z + vad_min*(1-z))
+end
+
+""
+function variable_ne_branch_voltage(pm::AbstractLPACCModel; kwargs...)
+    variable_ne_branch_voltage_magnitude_fr(pm; kwargs...)
+    variable_ne_branch_voltage_magnitude_to(pm; kwargs...)
+    variable_ne_branch_voltage_angle_fr(pm; kwargs...)
+    variable_ne_branch_voltage_angle_to(pm; kwargs...)
+    variable_ne_branch_cosine(pm; kwargs...)
+end
+
+""
+function variable_ne_branch_voltage_magnitude_fr(pm::AbstractLPACModel; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
+    buses = ref(pm, nw, :bus)
+    branches = ref(pm, nw, :ne_branch)
+
+    phi_fr_ne = var(pm, nw)[:phi_fr_ne] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :ne_branch)], base_name="$(nw)_phi_fr_ne",
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["f_bus"]), "phi_fr_start")
+    )
+
+    if bounded
+        for (i, branch) in ref(pm, nw, :ne_branch)
+            JuMP.set_lower_bound(phi_fr_ne[i], buses[branches[i]["f_bus"]]["vmin"] - 1.0)
+            JuMP.set_upper_bound(phi_fr_ne[i], buses[branches[i]["f_bus"]]["vmax"] - 1.0)
+        end
+    end
+
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :phi_fr, ids(pm, nw, :ne_branch), phi_fr_ne)
+end
+
+""
+function variable_ne_branch_voltage_magnitude_to(pm::AbstractLPACModel; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
+    buses = ref(pm, nw, :bus)
+    branches = ref(pm, nw, :ne_branch)
+
+    phi_to_ne = var(pm, nw)[:phi_to_ne] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :ne_branch)], base_name="$(nw)_phi_to_ne",
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["t_bus"]), "phi_to_start")
+    )
+
+    if bounded
+        for (i, branch) in ref(pm, nw, :ne_branch)
+            JuMP.set_lower_bound(phi_to_ne[i], buses[branches[i]["t_bus"]]["vmin"] - 1.0)
+            JuMP.set_upper_bound(phi_to_ne[i], buses[branches[i]["t_bus"]]["vmax"] - 1.0)
+        end
+    end
+
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :phi_to, ids(pm, nw, :ne_branch), phi_to_ne)
+end
+
+""
+function variable_ne_branch_voltage_angle_fr(pm::AbstractLPACModel; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
+    buses = ref(pm, nw, :bus)
+    branches = ref(pm, nw, :ne_branch)
+
+    va_fr_ne = var(pm, nw)[:va_fr_ne] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :ne_branch)], base_name="$(nw)_va_fr_ne",
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["f_bus"]), "va_fr_start")
+    )
+
+    # may be bounds can be added in future
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :va_fr, ids(pm, nw, :ne_branch), va_fr_ne)
+end
+
+""
+function variable_ne_branch_voltage_angle_to(pm::AbstractLPACModel; nw::Int=pm.cnw, bounded::Bool=true, report::Bool=true)
+    buses = ref(pm, nw, :bus)
+    branches = ref(pm, nw, :ne_branch)
+
+    va_to_ne = var(pm, nw)[:va_to_ne] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :ne_branch)], base_name="$(nw)_va_to_ne",
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["t_bus"]), "va_to_start")
+    )
+
+    # may be bounds can be added in future
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :va_to, ids(pm, nw, :ne_branch), va_to_ne)
+end
+
+
+function variable_ne_branch_cosine(pm::AbstractLPACCModel; nw::Int=pm.cnw, report::Bool=true)
+
+    bi_bp = Dict((i, (b["f_bus"], b["t_bus"])) for (i,b) in ref(pm, nw, :ne_branch))
+    buspair = ref(pm, nw, :ne_buspairs)
+    cs_ne = var(pm, nw)[:cs_ne] = JuMP.@variable(pm.model,
+        [b in ids(pm, nw, :ne_branch)], base_name="$(nw)_cs_ne",
+        lower_bound = 0,
+        upper_bound = 1,
+        start = comp_start_value(ref(pm, nw, :ne_buspairs, bi_bp[b]), "cs_start", 1.0)
+    )
+
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :cs_ne, ids(pm, nw, :ne_branch), cs_ne)
+end
+
+
+function constraint_ne_model_voltage(pm::AbstractLPACCModel, n::Int)
+    buses = ref(pm, n, :bus)
+    branches = ref(pm, n, :ne_branch)
+    bi_bp = Dict((i, (b["f_bus"], b["t_bus"])) for (i,b) in branches)
+    buspair = ref(pm, n, :ne_buspairs)
+    cos_min, cos_max  = ref_calc_angle_difference_bounds(ref(pm, n, :ne_buspairs))
+
+    phi  = var(pm, n, :phi)
+    z  = var(pm, n, :branch_ne)
+
+    cs = var(pm, n, :cs)
+    cs_ne = var(pm, n, :cs_ne)
+    phi_fr = var(pm, n, :phi_fr_ne)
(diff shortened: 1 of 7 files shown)


src/form/lpac.jl
```diff
@@ -130,3 +130,254 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    va_fr  = var(pm, n, :va_fr_ne, i)
+    va_to  = var(pm, n, :va_to_ne, i)
+    cs     = var(pm, n, :cs_ne, i)
+    z = var(pm, n, :branch_ne, i)
+    JuMP.@constraint(pm.model, p_to ==  (g+g_to)*(z + 2*phi_to) + (-g*tr-b*ti)/tm^2*(cs + phi_fr + phi_to) + (-b*tr+g*ti)/tm^2*-(va_fr-va_to) )
+    JuMP.@constraint(pm.model, q_to == -(b+b_to)*(z + 2*phi_to) - (-b*tr+g*ti)/tm^2*(cs + phi_fr + phi_to) + (-g*tr-b*ti)/tm^2*-(va_fr-va_to) )
+end
+
+
+function ref_calc_angle_difference_bounds(buspairs, conductor::Int=1)
```
ccoffrin, line 357: This function should be placed in `src/base/ref.jl`.  However, I don't think you need this function, see previous note.

src/form/lpac.jl
```diff
@@ -130,3 +130,254 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+
+    for (l,i,j) in ref(pm, n, :ne_arcs_from)
+        JuMP.@constraint(pm.model, phi_fr[l] <= z[l]*(buses[branches[l]["f_bus"]]["vmax"]-1))
+        JuMP.@constraint(pm.model, phi_fr[l] >= z[l]*(buses[branches[l]["f_bus"]]["vmin"]-1))
+
+        JuMP.@constraint(pm.model, cs_ne[l] <= z[l]*cos_max[bi_bp[l]])
+        JuMP.@constraint(pm.model, cs_ne[l] >= z[l]*cos_min[bi_bp[l]])
+
+        JuMP.@constraint(pm.model, phi_to[l] <= z[l]*(buses[branches[l]["t_bus"]]["vmax"]-1))
+        JuMP.@constraint(pm.model, phi_to[l] >= z[l]*(buses[branches[l]["t_bus"]]["vmin"]-1))
```
ccoffrin, line 267: For each of these, there is a simple solution,
```
a_lb, a_ub = _IM.variable_domain(a)
@assert (a_lb <= 0 && a_ub >= 0)
JuMP.@constraint(pm.model, a <= z[l]*a_ub)
JuMP.@constraint(pm.model, a >= z[l]*a_lb)
```
As long as the variables bounds are setup correctly this will work and will be a valid on-off constraint.

src/form/lpac.jl
```diff
@@ -130,3 +130,254 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :cs_ne, ids(pm, nw, :ne_branch), cs_ne)
+end
+
+
+function constraint_ne_model_voltage(pm::AbstractLPACCModel, n::Int)
+    buses = ref(pm, n, :bus)
+    branches = ref(pm, n, :ne_branch)
+    bi_bp = Dict((i, (b["f_bus"], b["t_bus"])) for (i,b) in branches)
+    buspair = ref(pm, n, :ne_buspairs)
+    cos_min, cos_max  = ref_calc_angle_difference_bounds(ref(pm, n, :ne_buspairs))
```
ccoffrin, line 245: This function can be removed once updates are made to the on-off variants here.

src/form/lpac.jl
```diff
@@ -130,3 +130,254 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    )
+
+    report && _IM.sol_component_value(pm, nw, :ne_branch, :cs_ne, ids(pm, nw, :ne_branch), cs_ne)
+end
+
+
+function constraint_ne_model_voltage(pm::AbstractLPACCModel, n::Int)
+    buses = ref(pm, n, :bus)
+    branches = ref(pm, n, :ne_branch)
+    bi_bp = Dict((i, (b["f_bus"], b["t_bus"])) for (i,b) in branches)
```
ccoffrin, line 243: Typically in NE models you cannot reason in terms of buspairs, becouse each line can be built individually you cannot reuse the variables that are shared between bus pairs.

src/form/lpac.jl
```diff
@@ -130,3 +130,254 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    JuMP.@constraint(pm.model, va_fr - va_to <= angmax*z + vad_max*(1-z))
+    JuMP.@constraint(pm.model, va_fr - va_to >= angmin*z + vad_min*(1-z))
+end
+
+""
+function variable_ne_branch_voltage(pm::AbstractLPACCModel; kwargs...)
+    variable_ne_branch_voltage_magnitude_fr(pm; kwargs...)
+    variable_ne_branch_voltage_magnitude_to(pm; kwargs...)
+    variable_ne_branch_voltage_angle_fr(pm; kwargs...)
+    variable_ne_branch_voltage_angle_to(pm; kwargs...)
```
ccoffrin, line 149: These constraints should reason over voltage angle differences `vad` and not specific angles, `va_fr` and `va_to`. 

See `variable_branch_voltage_product_angle_on_off` as an example, which you can likely reuse.

src/form/lpac.jl
ccoffrin: Slight correction here.  This will only work if the "on" bounds cross zero.  If they do not, then the values of `a_lb, a_ub` should be replaced with the "on" bounds.

src/form/lpac.jl
jay-dave: Thanks for the review @ccoffrin. I have a better idea of what needs to be done now. I would try to resolve them in a week or so.

ccoffrin:
No problem @jay-dave!  We may need a few iterations to get this sorted out, but I am really excited to see this feature on a path to getting into PowerModels core!

jay-dave:
@ccoffrin  In the new commit, I have tried doing both ots and tnep problem. The ots problem still needs some work since the objective values for case 3 and case 5 for LPAC is higher than AC. Also, the functions need to be arranged in an order. But I would like to check the basic formulations first and furnish it as we go along. It indeed seems like we will need a few iterations.

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+        sum(qg[g] for g in bus_gens)
+        - sum(qs[s] for s in bus_storage)
+        - sum(qd for qd in values(bus_qd))
+        + sum(bs for bs in values(bus_bs))*(1 + 2*phi)
+    )
+    if _IM.report_duals(pm)
+        sol(pm, n, :bus, i)[:lam_kcl_r] = cstr_p
+        sol(pm, n, :bus, i)[:lam_kcl_i] = cstr_q
+    end
+end
```
ccoffrin, line 378: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+
+""
+function constraint_ne_ohms_yt_from(pm::AbstractLPACCModel, n::Int, i, f_bus, t_bus, f_idx, t_idx, g, b, g_fr, b_fr, tr, ti, tm, vad_min, vad_max)
+    p_fr = var(pm, n,    :p_ne, f_idx)
+    q_fr = var(pm, n,    :q_ne, f_idx)
+    phi_fr = var(pm, n, :phi_fr_ne, i)
+    phi_to = var(pm, n, :phi_to_ne, i)
+    td = var(pm, n, :td_ne, i)
+    cs = var(pm, n, :cs_ne, i)
+    z = var(pm, n, :branch_ne, i)
```
ccoffrin, line 388: Nice trick. :-)

src/form/lpac.jl
ccoffrin: Function not used?  Can be removed.

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+""
+function variable_bus_voltage_on_off(pm::AbstractLPACCModel; kwargs...)
+    variable_bus_voltage_angle(pm; kwargs...)
+    variable_bus_voltage_magnitude(pm; kwargs...)
+
+    variable_branch_voltage_magnitude_fr_on_off(pm; kwargs...)
+    variable_branch_voltage_magnitude_to_on_off(pm; kwargs...)
+    variable_branch_voltage_product_angle_on_off(pm; kwargs...)
+    variable_branch_cosine_on_off(pm; kwargs...)
+end
```
ccoffrin, line 444: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+
+    phi_fr = var(pm, nw)[:phi_fr] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :branch)], base_name="$(nw)_phi_fr",
+        lower_bound = min(0, buses[branches[i]["f_bus"]]["vmin"] - 1.0),
+        upper_bound = max(0, buses[branches[i]["f_bus"]]["vmax"] - 1.0),
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["f_bus"]), "phi_fr_start")
+    )
+
+    report && _IM.sol_component_value(pm, nw, :branch, :phi_fr, ids(pm, nw, :branch), phi_fr)
+end
```
ccoffrin, line 460: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+
+    phi_to = var(pm, nw)[:phi_to] = JuMP.@variable(pm.model,
+        [i in ids(pm, nw, :branch)], base_name="$(nw)_phi_to",
+        lower_bound = min(0, buses[branches[i]["t_bus"]]["vmin"] - 1.0),
+        upper_bound = max(0, buses[branches[i]["t_bus"]]["vmax"] - 1.0),
+        start = comp_start_value(ref(pm, nw, :bus, branches[i]["t_bus"]), "phi_to_start")
+    )
+
+    report && _IM.sol_component_value(pm, nw, :branch, :phi_to, ids(pm, nw, :branch), phi_to)
+end
```
ccoffrin, line 475: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    q_fr = var(pm, n,    :q, f_idx)
+    phi_fr = var(pm, n, :phi_fr, i)
+    phi_to = var(pm, n, :phi_to, i)
+    td = var(pm, n, :td, i)
+    cs = var(pm, n, :cs, i)
+    z = var(pm, n, :z_branch, i)
+
+    JuMP.@constraint(pm.model, p_fr ==  (g+g_fr)/tm^2*(z + 2*phi_fr) + (-g*tr+b*ti)/tm^2*(cs + phi_fr + phi_to) + (-b*tr-g*ti)/tm^2*(td))
+    JuMP.@constraint(pm.model, q_fr == -(b+b_fr)/tm^2*(z + 2*phi_fr) - (-b*tr-g*ti)/tm^2*(cs + phi_fr + phi_to) + (-g*tr+b*ti)/tm^2*(td))
+end
```
ccoffrin, line 489: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    q_to = var(pm, n,    :q, t_idx)
+    phi_fr = var(pm, n, :phi_fr, i)
+    phi_to = var(pm, n, :phi_to, i)
+    td = var(pm, n, :td, i)
+    cs = var(pm, n, :cs, i)
+    z = var(pm, n, :z_branch, i)
+
+    JuMP.@constraint(pm.model, p_to ==  (g+g_to)*(z + 2*phi_to) + (-g*tr-b*ti)/tm^2*(cs + phi_fr + phi_to) + (-b*tr+g*ti)/tm^2*-(td) )
+    JuMP.@constraint(pm.model, q_to == -(b+b_to)*(z + 2*phi_to) - (-b*tr+g*ti)/tm^2*(cs + phi_fr + phi_to) + (-g*tr-b*ti)/tm^2*-(td) )
+end
```
ccoffrin, line 503: 👍

src/form/lpac.jl
```diff
@@ -130,3 +131,373 @@ function constraint_ohms_yt_to(pm::AbstractLPACCModel, n::Int, f_bus, t_bus, f_i
...
+    JuMP.@constraint(pm.model, va_fr - va_to >= angmin*z + vad_min*(1-z))
+end
+
+""
+function variable_ne_branch_voltage(pm::AbstractLPACCModel; kwargs...)
+    variable_ne_branch_voltage_magnitude_fr(pm; kwargs...)
+    variable_ne_branch_voltage_magnitude_to(pm; kwargs...)
+    variable_ne_branch_voltage_angle(pm; kwargs...)
+    variable_branch_cosine_ne(pm; kwargs...)
+    variable_buspair_voltage_product_angle(pm; kwargs...)
```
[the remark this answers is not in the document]
jay-dave: I guess that's right, only one variable will do the trick. The angle difference constraint  ```constraint_ne_model_voltage``` becomes non-binding for the branches that are not built. That's the point I missed and used two variables ```:td``` and ```:td_ne``` for parallel branches, assuming angle difference ```:td``` will be zero when branch is not built. But that's unnecessary.

jay-dave:
@ccoffrin Thanks for the review.

Along with other minor updates, OTS test cases are included in this commit. I am not sure why there are different set of tests for different formulations in ```ots.jl```, but I guess it has to do with checking a particular aspect of the formulations. For now, I have copied tests from QC along with their set ```atol``` (=50 in one test) since this formulation is near to LPAC. As you suggested, I could observe some lines switching off to reduce the objective value from its OPF case so our implementation seems to be working.

I have used some functions i.e. ```variable_branch_cosine_on_off``` and ```variable_branch_voltage_product_angle_on_off``` from ```wr.jl``` for LPAC. I am not sure if they need to be moved to ```shared.jl``` or used in ```lpac.jl```in any other way.

Feel free to suggest more corrections if required.

ccoffrin:
Thanks @jay-dave, the changes you made look good to me.  I am going to do one last pass from scratch and will merge after that.  This will get into the next PM release.

PR #758, opened 2021-01-25
In-place data comp. #721

### LINKED ISSUE #721: In-place Data Computations
Make variants of `calc_branch_flow_ac`,`calc_branch_flow_dc`,`calc_power_balance` that update the given data in-place.

jbarberia:
To close #721

### DIFF
--- a/src/core/data.jl
+++ b/src/core/data.jl
@@ -764,6 +764,13 @@ function calc_branch_flow_ac(data::Dict{String,<:Any})
 end
 
 
+"In place calculation calc_branch_flow_ac"
+function calc_branch_flow_ac!(data::Dict{String,<:Any})
+	flows = calc_branch_flow_ac(data)
+	update_data!(data, flows)
+end
+
+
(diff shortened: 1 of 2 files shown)


ccoffrin:
Thanks for the contribution.  This issue was not particularly well articulated but one of the objectives was to avoid allocating memory (which is one of the more time consuming aspects of this).  The idea would be to make variants of `calc_branch_flow_ac` that update `data` directly instead of making a separate dict that can be merged with `update_data`

That said, I am open to merging this as is as a convince function where the implementation may change later.  If so, we should add a new issue along the lines of "avoid memory allocation in functions ..." to remember to update the implementation.

jbarberia:
Sorry for the misunderstanding, I thought it was just implementing the function. At the moment I do not plan to work on improving the memory allocation of this functions.

But it would be a good thing to update issue #721 to indicate that memory allocation needs to be improved.

Feel free to close or merge this pull request.

ccoffrin:
Ok, I will close this for now and update #721.

PR #763, opened 2021-02-07
Add Symmetric DCLL Formulation

andrewrosemberg:
Add Symmetric DCLL Formulation that divides losses both ways.

### DIFF
--- a/test/opf.jl
+++ b/test/opf.jl
@@ -438,6 +438,32 @@ end
     end
 end
 
+@testset "test dc+ll symmetric opf" begin
+
+    @testset "5-bus case" begin
+        result = run_opf("../test/data/matpower/case5.m", SymmetricDCPLLPowerModel, ipopt_solver)
+
+        @test result["termination_status"] == LOCALLY_SOLVED
+        @test isapprox(result["objective"], 17765; atol = 1e0)
+        @test isapprox(result["solution"]["branch"]["1"]["pt"], -2.40978; atol = 1e-3)
+        @test isapprox(result["solution"]["branch"]["1"]["pf"], 2.39357; atol = 1e-3)
+    end
+
+    data = PowerModels.parse_file("../test/data/matpower/case5.m")
+    new_f_bus = data["branch"]["1"]["t_bus"]
+    data["branch"]["1"]["t_bus"] = data["branch"]["1"]["f_bus"]
+    data["branch"]["1"]["f_bus"] = new_f_bus
+
+    @testset "5-bus reverse case" begin
+        result = run_opf(data, SymmetricDCPLLPowerModel, ipopt_solver)
+
+        @test result["termination_status"] == LOCALLY_SOLVED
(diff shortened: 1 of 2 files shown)


ccoffrin:
@andrewrosemberg, thanks for your proposal!   I was having a conversation about this with @joaquimg some time ago so I am curious to hear his two cents as well.

I had a quick look at what the AC model does and the active power loss appears to be always one-sided. That is, the side of the branch that consumes active power (i.e. input) is greater than the side that produces (i.e. output), this seems intuitive given the idea of "line losses".

I then compared this to the solutions I get when running Ipopt on the original (non-symmetrical) DC+loss model at it seems to produce the same behavior.  I have not thought about it deeply why the original formulation has this property, but it does seem to be advantageous in the spirit of modeling losses in a way that is similar to the AC model.

My first question about this proposal is what is the motivation to model the line losses as symmetrical?

joaquimg:
Hey!
Yes, I did talk to @andrewrosemberg , and I am very interested in this.

First, I think this should be an alternate formulation instead of a replacement. This would be very aligned with the spirit of PowerModels.

My motivations are:
 - This is a formulation I have been using for a long time at work
 - The symmetric formulation is label independent, with that I mean that, in the asymmetric form, changing labels/order of buses changes results.
 
 Moreover, there might be literature gaps on this subject, I'd be happy to collaborate on comparing formulation and how well they match the AC model.

andrewrosemberg:
Good points! I have added a test using this case (`case5`) so we can analyze the performance.

For some reason, I'm getting an error due to some sort of asymmetry.

andrewrosemberg:
Ready for a review :)

ccoffrin:
The implementation looks reasonable to me but I @joaquimg would know best if this is a faithful replication of the symmetrical model he had used in the past.

Once he has signed off on the implementation, I request that we move this feature to a PR on the PowerModelsAnnex package and close this one.

test/opf.jl
```diff
@@ -438,6 +438,32 @@ end
...
+        @test result["termination_status"] == LOCALLY_SOLVED
+        @test isapprox(result["objective"], 17765; atol = 1e0)
+        @test isapprox(result["solution"]["branch"]["1"]["pt"], -2.39357; atol = 1e-3)
+        @test isapprox(result["solution"]["branch"]["1"]["pf"], 2.40978; atol = 1e-3)
+    end
+
+    data = PowerModels.parse_file("../test/data/matpower/case5.m")
+    new_f_bus = data["branch"]["1"]["t_bus"]
+    data["branch"]["1"]["t_bus"] = data["branch"]["1"]["f_bus"]
+    data["branch"]["1"]["f_bus"] = new_f_bus
```
joaquimg, line 455: Maybe move this block inside the tests

andrewrosemberg: moved

ccoffrin:
closing due to lack of activity.  A PR to PowerModelsAnnex is recommended for any future work.

PR #766, opened 2021-02-18
building expression in two steps 

claytonpbarrows:
This PR resolves the following error when building a voltage angle expression with terms that are `::ParameterRef`:

```
ArgumentError: Cannot call `mutable_operate!(add_mul, ::ParameterJuMP.DoubleGenericAffExpr{Float64,VariableRef,ParameterRef}, ::Float64, ::ParameterJuMP.DoubleGenericAffExpr{Float64,VariableRef,ParameterRef})` as objects of type `ParameterJuMP.DoubleGenericAffExpr{Float64,VariableRef,ParameterRef}` cannot be modifed to equal the result of the operation. Use `operate!` instead which returns the value of the result (possibly modifying the first argument) to write generic code that also works when the type cannot be modified.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,8 +31,9 @@ julia = "^1"
 Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
 Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
 Juniper = "2ddba703-00a4-53a7-87a5-e8b9971dde84"
+ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Ipopt", "SCS", "Cbc", "Juniper"]
+test = ["Test", "Ipopt", "SCS", "Cbc", "Juniper", "ParameterJuMP"]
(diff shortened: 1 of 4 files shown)


ccoffrin:
Thanks @claytonpbarrows.  I have a vague recollection that there is a performance or robustness issue if we do not use the `@expression` macro, @odow can you share your two cents?

In either case, we should develop a test in PowerModels that exhibits this error so that we do not have a regression after we have resolved this point.

ccoffrin:
Thanks for the test Clayton.  

@odow or @blegat can I get your input on the pros/cons of using the `@expression` macro or not?

blegat:
If it's the sum of a vector there is no difference but if it's the sum of a Generator like here, it would have a big difference, you need to use `MA.@rewrite` or `JuMP.@expression`
https://github.com/jump-dev/MutableArithmetics.jl/blob/595460dd664380e7668624fccb4f26a637fc1af6/src/dispatch.jl#L9

odow:
Is this a bug in ParameterJuMP then? The `@expression` syntax should work, and it is the preferred option.

ccoffrin:
@joaquimg, thoughts on if this should be fixed at the ParameterJuMP level?

joaquimg:
It seems that we need to define the method in ParameterJuMP indeed.

ccoffrin:
Based on this discussion, I am inclined to say this is a bug that should be addressed in ParameterJuMP and we should not make the proposed change in PowerModels.  Hopefully @joaquimg can adapt the test @claytonpbarrows made into ParameterJuMP.

claytonpbarrows:
@joaquimg, please let me know how you would like to proceed. If you'd prefer, I can make an attempt at a PR, though I may need some guidance.

claytonpbarrows:
closing this with the preference to address the issue in ParameterJuMP

joaquimg:
You can open an issue or PR in ParameterJuMP and we go from there.
The error you reported highlights the missing function implementation. Sure I can help.

jd-lara:
@joaquimg and @blegat I think this is related to https://github.com/JuliaStochOpt/ParameterJuMP.jl/issues/66

Issue #762, opened 2021-02-03
Jacobian Support for Basic Network Data


ccoffrin:
A Jacobian matrix support to the Basic Network Data features.  Key features include,
- building a from-scratch AC Power Flow solver on basic data
- analysis p/q vs vm/theta sensitive (i.e. Jacobian were all buses are PQ)
- from-scratch fast decoupled load flow solver (optional)

jbarberia:
@ccoffrin, I started working to contribute to this issue. Tell me what you think about [this](https://github.com/jbarberia/PowerModels.jl/commit/1595190e54c10aa41638bc3c9c84a928db59e0df).

I know it can be improved (specially the AC Power Flow solver). Surely this week I can finish it and make a pull request.

ccoffrin:
@jbarberia thanks for taking a pass at this.  I only had a quick pass but here are some initial comments,

- I would prefer to not have a function with the signature `calc_basic_jacobian_matrix(data::Dict{String,<:Any}, V::Array{Complex{Float64}, 1})`, a user can change the bus type in the dict before sending it to `calc_basic_jacobian_matrix`
- I would move the `calc_basic_bus_type` code into `calc_basic_jacobian_matrix` I am not sure it has many uses outside of that function.
- I would move the `compute_basic_ac_pf` code into the package tests.  Users who would like to solve AC-PF (on basic data or not) should use `run_ac_pf` or `compute_ac_pf`.

I'll also loop in @molzahn as this was a feature he requested.

ccoffrin:
closed by #777

PR #777, opened 2021-05-26
Jacobian Support for Basic Network

jbarberia:
#762

I modified the commented out version of the function that calculates the Jacobian matrix for basic networks. Also add a function to compute the sub-matrices dP/dva and dQ/dvm.

I move the from scratch AC power flow solver to `test/common.jl` because the Jacobian matrix is tested with it.

### DIFF
--- a/test/data-basic.jl
+++ b/test/data-basic.jl
@@ -189,39 +189,61 @@ end
         end
     end
 
-    @testset "test basic jacobian" begin  
+    @testset "test basic jacobian" begin
+        @testset "5-bus-case" begin
+            data = make_basic_network(PowerModels.parse_file("../test/data/matpower/case5.m"))
+            for (i, bus) in data["bus"] # All buses PQ
+                bus["bus_type"] = 1
+            end
+            J = PowerModels.calc_basic_jacobian_matrix(data);
+
+            num_bus = length(data["bus"])
+            
+            @test size(J)[2] == num_bus * 2
+            @test size(J)[1] == num_bus * 2
+            @test isapprox(LinearAlgebra.det(J), 1524.5588; atol=1e-4)
+            
+            # PQ bus - diagonal
+            @test isapprox(J[1, 1], 225.5254; atol=1e-4)                  # dP/dva diagonal
+            @test isapprox(J[num_bus+1, 1], -20.7074; atol=1e-4)          # dQ/dva diagonal
+            @test isapprox(J[1, num_bus+1], 23.7940; atol=1e-4)           # dP/dvm diagonal
+            @test isapprox(J[num_bus+1, num_bus+1], 219.4434; atol=1e-4)  # dQ/dvm diagonal
+        end
+        
         @testset "24-bus-case" begin
             data = make_basic_network(PowerModels.parse_file("../test/data/matpower/case24.m"))
             J = PowerModels.calc_basic_jacobian_matrix(data)
             
             num_bus = length(data["bus"])
-
+            
             @test size(J)[1] == num_bus * 2
             @test size(J)[2] == num_bus * 2
+            @test isapprox(LinearAlgebra.det(J), 3.133520767308452e+57)
+            @test isapprox(sum(J), -3.5289; atol=1e-4)
 
             # PV bus - diagonal
-            @test isapprox(J[1, 1], 90.2500; atol = 1e-4)                   # dP/dva diagonal
-            @test isapprox(J[num_bus+1, 1], -17.2488; atol = 1e-4)          # dQ/dva diagonal
-            @test isapprox(J[1, num_bus+1], 0.0; atol = 1e-4)               # dP/dvm diagonal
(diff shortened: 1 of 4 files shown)


ccoffrin:
After another quick review, nothing stands out to me as notably problematic.  I might rename some of the local variables to be more inline with the code conventions but that is a minor point as long as the code is correct.

Some requests regarding the unit tests,
- The test for `calc_basic_jacobian_matrix` should include a case where a network is loaded in, all buses are changed to p/q and then `calc_basic_jacobian_matrix` is called.  I suggest using `case5.m` as the base network. 
- All of the `calc_basic_jacobian_matrix` tests should include some metric that captures the full matrix, so that we can detect changes that do not occur in specific cells. `LinearAlgebra.det` seems to work well for the admittance matrix tests.
- We need to cross validate the results with what comes out of Matpower.  If you are not able to do this I can help.

jbarberia:
The suggestions made seem fine to me.

* I like the idea of testing in a case with all p/q buses.
* I'm going to add the determinant in the tests.
* How would the correlation with MATPOWER be?
[Here](https://github.com/lanl-ansi/PowerModels.jl/pull/777/commits/b5611b9b960f28e7fd90a422d3f45f290e824976#diff-014eb8a32b127fcf3633944b967ee8a2f0f26c56d53cbe4cb8e9db76d31f966eR193) I simply took the values of the Jacobian made in MATPOWER.

Tomorrow I can add a commit whit this.

ccoffrin:
As long as you did a side by side check with the output of Matpower's output, then we are good.

jbarberia:
@ccoffrin, I had a problem with the unit test for the case 5:

The workflow that I do the side by side comparison with Matpower was:
```julia
using PowerModels
data = make_basic_network(PowerModels.parse_file("../test/data/matpower/case5.m"))
for (i, bus) in data["bus"] # All buses PQ
    bus["bus_type"] = 1
end
export_to_matpower("case5basic.m", data)
J = calc_basic_jacobian_matrix(data)
```
and then make the Jacobian with Matpower:
```matlab
J = full(makeJac((case5basic), 1)) % The 1 is for the full jacobian
```

It was strange to notice that the determinant of the Jacobian in Julia and in Matlab gave very different values ​​for case 5. Even if the matrices were the same!

I checked it using the `MAT` module to load the Jacobian matrix from Matpower to Julia
```julia
using MAT
J_matpower = matread("J.mat")["J"]
for i in 1:10
    for j in 1:10
        @test isapprox(J[i, j], J_matpower[i, j])
    end
end 
```

test/data-basic.jl
```diff
@@ -189,53 +189,74 @@ end
...
+            for (i, bus) in data["bus"] # All buses PQ
+                bus["bus_type"] = 1
+            end
+            J = PowerModels.calc_basic_jacobian_matrix(data);
+
+            num_bus = length(data["bus"])
+            
+            @test size(J)[2] == num_bus * 2
+            @test size(J)[1] == num_bus * 2
+            @test isapprox(LinearAlgebra.det(J), 1524.5588; atol=1e-4)
```
jbarberia, line 204: The `LinearAlgebra.det` is not a good full matrix metric.
``` julia
J = calc_basic_jacobian_matrix(data) # J :: SparseMatrixCSC
J |> det # 1524.55...
J |> Matrix |> det # -1590.68...
U, V = svd(J |> Matrix); reduce(*, V) # 3967.00...
```

What do you think about something simple like:
``` julia
J = calc_basic_jacobian_matrix(data) # J :: SparseMatrixCSC
metric(arr) = reduce(*, [sum(row) for row in eachcol(arr)])
J |> metric # -132063.9683811387
J |> Matrix |> metric # -132063.9683811387
```

ccoffrin: I an fine with the sum metric instead.

jbarberia:
@ccoffrin, Could you review the code to be able to merge the PR?

jbarberia:
@ccoffrin, Should I close this PR?

ccoffrin:
No way!  Sorry for the delay, I do plan to merge this hopefully before the next patch release.

Were you in a hurry to get it to be part of the PowerModels release?

jbarberia:
Don't worry. I ain't hurry with this PR. I only wanted to know if it was useful to leave it open or not.

ccoffrin:
Hi @jbarberia concerning the fast decoupled solver, I like the implementation that you propose because it is algorithmically the same as the fully coupled case.  However, as I recall, the traditional fast decoupled load flow has a structure of update, Theta, recompute, update V, recompute, update Theta, ...; rather than update Theta/V, recompute, update Theta/V, ...

Do you know if there is some president/reference for the approach you have taken here?

ccoffrin:
@jbarberia, can you add yourself to the README.md file as a contributor of the basic network Jacobean feature?

jbarberia:
@ccoffrin, for sure!

PR #788, opened 2021-09-12
aligning 3-bus areas with area definition

claytonpbarrows:
The case3.m test file has area definitions where `pm_data["areas"]["2"]["col_2"] == 3` suggesting that the reference bus is bus 3, but `pm_data["bus"]["3"]["area"] == 1`. This fixes the discrepancy in the data file. If the mismatch is intentionally handled by PM, then please disregard. If it's actually a bug that should be caught, I can add a simple test.


ccoffrin:
The `areas` data matrix is not a pre-defined data source by PowerModels. It falls under the category of ad-hoc data extensions.  So in this case, PowerModels would assume that the bus data is correct and ignore this areas block.

I have no problem making these two parts of the data consistent, but in general would suggest using the bus-data for determining what areas exist in the data.  Does that sound reasonable?  Maybe you can tell me a bit more about your use case.

claytonpbarrows:
I built [this](https://github.com/NREL-SIIP/PowerModelsInterface.jl/tree/cb/pm_translator) that builds a (mostly) complete PowerModels dict from a `PowerSystems.System`. The issue is just that my test suite relies on the PowerModels test data, where in one of the tests I'm checking the consistency between a data set built directly from power models vs. one built from PowerSystems -> PowerModelsInterface -> PowerModels. In the latter, the `pmi_data["areas"]["2"]["col_2"]` is populated with options that are actually defined with an Area-2 membership in the `System`. So, effectively this process is just catching a data mismatch in case3.m. If you (understandably) don't care to enforce the consistency, I can either avoid the test, or we can merge this data fix.

Issue #754, opened 2021-01-19
Export HVDC to psse


jbarberia:
Add support to HVDC in the `export_pti` function.

Issue #540, opened 2019-06-10
Improved HVDC Line Model


ccoffrin:
It would be ideal to have a model similar to our current storage model.  Should focus on the two-terminal model and level multi-node models to PowerModelsACDC.

ccoffrin:
@hakanergun proposed to rename `dcline` to `dcbranch`

PR #791, opened 2021-09-24
Export HVDC to PSSE (fix: #754)

jbarberia:
What do you think of having a support to export HVDC lines in this way?

I compare side by side the results in PSSE  doing this process: `original_raw -> parse_file -> export_file -> new_raw` and its works pretty decent.

The only drawback is that it would work correctly if `import_all = true` in the `parse_file function`. 
For the moment I did not check if it could export to PSSE a file coming from MATPOWER.


ccoffrin:
I did a quick skim and this is how I would approach it as well.  If the user does `import_all` then the export feature can recover more data than is available in the current PowerModels spec. I would strive to for the goal that `original_raw -> parse_file (import_all) -> export_file -> new_raw` yields the exact same DC line record in the exported PSSE file (for the two terminal case).  This may not always be possible if information is lost in translation to the power models data fields.

Note
- We should have a test for `original_raw -> parse_file (not import_all) -> export_file -> new_raw`, to make sure this does not crash and produces a reasonable output.
- We should have a test for `matpower (with DC line) -> parse_file -> export_file -> new_raw`, to make sure this does not crash and produces a reasonable output.

jbarberia:
I agree with you to add this types of tests. In the next days I will try to produce a useful output without the `import_all` option.

jbarberia:
@ccoffrin, I would like to know what you think about the power flows that I ran for `case5_dc.m`.

- First, I solved it in matpower `runpf(loadcase("case5_dc.m"))`.
- Then in PowerModels `run_pf("case5_dc.m", ACPPowerModel, Ipopt.Optimizer)` .
- Lastly, I used the `export_pti` function and ran a powerflow with default options in `PSSE-33`. 

The results are in the following tables:

Voltage Magnitude [p.u.] | PowerModels | Matpower | PSSE-33
-- | -- | -- | --
1 | 1.064 | 1.076 | 1.078
2 | 1.080 | 1.083 | 1.084
3 | 1.100 | 1.099 | 1.100
4 | 1.064 | 1.064 | 1.064
5 | 1.053 | 1.066 | 1.069

Voltage Angle [degree] | PowerModels | Matpower | PSSE-33
-- | -- | -- | --
1 | 2.910 | 2.808 | 2.830
2 | -0.768 | -0.683 | -0.750
3 | -0.622 | -0.486 | -0.600
4 | 0.000 | 0.000 | 0.000
5 | 3.734 | 3.585 | 3.630

Let me know what things would be necessary (more tests or another side by side comparision) to perform the merge.

jbarberia:
@ccoffrin, I prefer to merge this PR and then work on #540 to make all the necessary corrections there.

jbarberia:
@ccoffrin, I used the `two-terminal-hvdc_test_2.raw` file to verify that the `PSSE-33` could open the `.raw` cases generated by PowerModels. Also, I used it for some side-by-side checks. 

But now it is totally unnecessary, so I will delete it.

PR #792, opened 2021-10-09
add z i load conversion

jd-lara:
Handles the addition of load data when the raw file include impedance and current loads.

### DIFF
--- a/test/psse.jl
+++ b/test/psse.jl
@@ -515,7 +515,7 @@ end
     end
 
      @testset "Impedance and Current Load Conversions" begin
-        data = PowerModels.parse_file("/Users/jdlara/.julia/dev/PowerModels/test/data/pti/parser_test_k.raw")
+        data = PowerModels.parse_file("../test/data/pti//parser_test_k.raw")
         Vm_load = data["bus"]["103"]["vm"]
         # Test current load only
         @test isapprox(data["load"]["2"]["pd"], 250*Vm_load/data["baseMVA"], atol = 1e-6)


jd-lara:
@ccoffrin it seems that just the windows test is stuck loading Julia packages.

ccoffrin:
I see it. This part of the feature is looking good to me, I am also thinking about these two points,

1. I am wondering what the implications are for the `export_pti` feature are for, ideally it should un-do any changes to the constant power loads when doing the export when the `import_all` flag is set.
2. I am debating in my mind if this should be considered a feature add or a breaking change so I can merge into the appropriate branch. I am leaning to a breaking change at this point.

@jbarberia, you seem to be an active PSSE user and you developed the `export_pti` feature, what are your thoughts on the two points above?

jd-lara:
> I see it. This part of the feature is looking good to me, I am also thinking about these two points,
> 
> 1. I am wondering what the implications are for the `export_pti` feature are for, ideally it should un-do any changes to the constant power loads when doing the export when the `import_all` flag is set.
> 2. I am debating in my mind if this should be considered a feature add or a breaking change so I can merge into the appropriate branch. I am leaning to a breaking change at this point.
> 
> @jbarberia, you seem to be an active PSSE user and you developed the `export_pti` feature, what are your thoughts on the two points above?

I think it depends on the scope of the features you are thinking about. If your intention of `export_pti` is to enable roundtrip PSSe data conversion then it is breaking. If the intention of `export_pti` is to write PowerModels data as .raw files then it doesn't break anything since PowerModels data model doesn't have explicit entries for ZIP loads.

PR #801, opened 2021-11-29
Fix Sort Deprecation Warning

ccoffrin:
@jbarberia an FYI just for your general knowledge.

When running the package tests I was getting this warning,
```
Testing Running tests...
┌ Warning: `sort(d::Dict; args...)` is deprecated, use `sort!(OrderedDict(d); args...)` instead.
│   caller = ip:0x0
└ @ Core :-1
```
Interestingly this only shows up from inside the package manager, i.e. `] test PowerModels`.

I tracked it down to the calls in the `pti_export` feature like this one, `sort(data["bus"], by = (x) -> parse(Int64, x))`.

There are many possible fixes to the issue but the simple one in this situation is to "collect" the dict elements into an array before sorting them.

Additionally, a PowerModels specific tip, we have a convention that every component has a field called "index" that is the integer version of the component's id.  Using this we can avoid the parsing of the dict key in many cases.  In this case `by = (x) -> parse(Int64, x.first)` can be replaced by `by = (x) -> x.second["index"]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- nothing
+- Made case name recovery optional in PTI parsing
+- Fixed Julia deprecation warning when calling sort on Dict
 
 ### v0.19.1
 - Add support for Memento v1.3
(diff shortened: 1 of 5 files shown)


jbarberia:
@ccoffrin , Thanks for the information!
I will take it into account for future contributions.

PR #809, opened 2022-03-02
Updates for JuMP v0.23

ccoffrin:
minimal updates to get code working in JuMP v0.23


odow:
Let me take a pass at this.

PR #810, opened 2022-03-02
Update Project [compat]

odow:
I think you can do this as a non-breaking release.

You _might_ have someone complain that the tests are broken on Julia 1.0, but PowerModels should work.

And we don't explicitly test JuMP v0.22, but you didn't change anything so there shouldn't be an issue. We also don't test earlier versions of InfrastructureModels or JSON, but there's no issue including older compats with them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,10 @@ PowerModels.jl Change Log
 ### Staged
 - nothing
 
-### v0.20.0
+### v0.19.3
+
 - Add support for JuMP v0.23
-- Update minimum Julia requirement v1.6 (LTS)
 - Replace CBC with HiGHS in tests
-- Updates for SCS v0.9
 
 ### v0.19.2
 - Rename `run_*` methods to `solve_*` and add depreciation warnings (#649)
--- a/Project.toml
+++ b/Project.toml
@@ -14,15 +14,15 @@ NLsolve = "2774e3e8-f4cf-5e23-947b-6d7e65073b56"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-HiGHS = "1"
+HiGHS = "~0.3, 1"
 InfrastructureModels = "~0.6, ~0.7"
-Ipopt = "1"
+Ipopt = "~0.8, ~0.9, ~1"
 JSON = "~0.18, ~0.19, ~0.20, ~0.21"
 JuMP = "~0.22, ~0.23"
 Juniper = "~0.8, ~0.9"
 Memento = "1"
 NLsolve = "4"
-SCS = "1"
+SCS = "~0.8, ~0.9, ~1"
 julia = "1"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "PowerModels"
 uuid = "c36e90e8-916a-50a6-bd94-075b64ef4655"
 authors = ["Carleton Coffrin"]
 repo = "https://github.com/lanl-ansi/PowerModels.jl"
-version = "0.20.0"
+version = "0.19.3"
 
 [deps]
 InfrastructureModels = "2030c09a-7f63-5d83-885d-db604e0e9cc0"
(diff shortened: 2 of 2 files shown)


ccoffrin:
A a fresh Julia 1.0 install on my system I get this error,
```
(v1.0) pkg> test PowerModels
   Testing PowerModels
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package HiGHS [87dc4568]:
 HiGHS [87dc4568] log:
 ├─possible versions are: [0.1.0-0.1.3, 0.2.0-0.2.3, 0.3.0-0.3.2, 1.0.0] or uninstalled
 ├─restricted to versions * by an explicit requirement, leaving only versions [0.1.0-0.1.3, 0.2.0-0.2.3, 0.3.0-0.3.2, 1.0.0]
 └─restricted by julia compatibility requirements to versions: uninstalled — no versions left
```
But should be fine on Julia v1.6, so I am going to try restoring that lower bound and see if the package register will let it through as non-breaking.  Will see.

odow:
> On a fresh Julia 1.0 install on my system I get this error

Yes, but only if you run the tests. You should be able to _use_ PowerModels without issue.

ccoffrin:
Ya I get it, but I am not keen to some released version of PowerModels were `] test PowerModels` is known to fail.  Maybe the Julia package manager will let this lower bound change thorough with a patch, if not, I am fine with a "fake" breaking release to make things work.

odow:
How many people are still using Julia 1.0?

ccoffrin:
I would guess not many, so I am not too concerned about how this ends up getting resolved. I do care that tests pass on installable configurations of PowerModels.

odow:
The other factor to consider is that JuMP 1.0 is pretty close, so you can make a breaking PowerModels 1.0 release soon anyway.

ccoffrin:
I suppose you already know but this compat does not work on Julia v1.6 and JuMP v0.22 (details below).  But can work if the solver bounds are relaxed as I had them `SCS = ">= 0.9"`, `HiGHS = ">= 0.3"`
```
(@v1.6) pkg> test PowerModels
     Testing PowerModels
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:1531
ERROR: Unsatisfiable requirements detected for package MathOptInterface [b8f27783]:
 MathOptInterface [b8f27783] log:
 ├─possible versions are: 0.5.0-1.1.0 or uninstalled
 ├─restricted by compatibility requirements with JuMP [4076af6c] to versions: 0.10.4-0.10.9
 │ └─JuMP [4076af6c] log:
 │   ├─possible versions are: 0.18.3-0.23.0 or uninstalled
 │   ├─restricted to versions 0.22-0.23 by PowerModels [c36e90e8], leaving only versions 0.22.0-0.23.0
 │   │ └─PowerModels [c36e90e8] log:
 │   │   ├─possible versions are: 0.19.3 or uninstalled
 │   │   └─PowerModels [c36e90e8] is fixed to version 0.19.3
 │   └─restricted to versions 0.22.3 by an explicit requirement, leaving only versions 0.22.3
 └─restricted by compatibility requirements with SCS [c946c3f1] to versions: 1.0.0-1.1.0 — no versions left
   └─SCS [c946c3f1] log:
     ├─possible versions are: 0.4.0-1.0.1 or uninstalled
     └─restricted to versions 1 by an explicit requirement, leaving only versions 1.0.0-1.0.1
```

odow:
I set the compat bounds so that your CI would run, and under the assumption that we didn't care if the tests fail on older installs, even if PowerModels will work. But if you're going to make a lower bound of 1.6, then we can relax them.

odow:
How's this now?

ccoffrin:
SCS can only go down to 0.9 due to how the underlying solver changed, but I think this will work, testing locally on my branch.

ccoffrin:
Changes where minor so I made them locally for testing.  Seemed to work, here is what I came up with, https://github.com/lanl-ansi/PowerModels.jl/pull/809/commits/d9364f8832f74b803b231a40e9acde311cf5c25d

PR #834, opened 2022-07-30
Make SDP Solution Reporting Consistent

### LINKED ISSUE #833: SDP Solution Consistency
Currently `SDPWRMPowerModel` and `SparseSDPWRMPowerModel` provide different solution structures.  Should look into ways to make these more consistent.  For example, by constructing a sparse WI and WR matrices from the `w_group` values.

The other topic to consider is putting the `w,wi,wr` values into the bus and branch dicts or not.  

Discovered by @BabakT1

ccoffrin:
Add `w` values to buses and `wr,wi` to branches in all model variants.

Producing a consistent `W` matrix in all cases could be quite time consuming and memory intensive defeating the objective of the sparse solver.  Reporting these minimal values provides a user friendly solution.

Closes #833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- nothing
+- Fix SDP solution reporting consistency (#833)
 
 ### v0.19.6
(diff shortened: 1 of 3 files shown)


ccoffrin:
@BabakT1 this should address the issue we were discussing in #381

PR #839, opened 2022-08-20
Fix Tests on M1

ccoffrin:
@odow, these appear to be the only changes that need to be made. Appears to be a slight numerical difference in Julia's builtin `det` function.  I am going to wait to merge to get your input to do a work around or not in this case.


odow (approved):
This seems unrelated to M1. It could happen on any machine with different BLAS etc.

odow:
FWIW, I just reproduced these two failures as well.

PR #849, opened 2023-01-05
Faster PTDF matrix computation

mtanneau:
The current code for computing PTDF matrices (`calc_basic_ptdf_matrix`) relies on a dense matrix inverse, plus some loops that can be optimized.

This PR uses a more efficient, sparse factorization together with linear algebra operations to speed up the computation.

Performance comparison on some medium-size PGLib networks:

| casefile | before (s) | after (s) | speedup |
|:-------|--:|--:|--:|
| 1354_pegase              |     0.584 |     0.056 |    10.41 |
| 1888_rte                 |     1.781 |     0.132 |    13.51 |
| 1951_rte                 |     2.205 |     0.225 |     9.78 |
| 2000_goc                 |     1.817 |     0.117 |    15.58 |
| 2312_goc                 |     2.450 |     0.205 |    11.97 |
| 2383wp_k                 |     2.483 |     0.242 |    10.27 |
| 2736sp_k                 |     3.088 |     0.291 |    10.63 |
| 2737sop_k                |     3.349 |     0.309 |    10.83 |
| 2742_goc                 |     4.303 |     0.325 |    13.24 |
| 2746wop_k                |     3.051 |     0.292 |    10.46 |
| 2746wp_k                 |     3.505 |     0.286 |    12.24 |
| 2848_rte                 |     3.628 |     0.318 |    11.40 |
| 2853_sdet                |     4.077 |     0.311 |    13.09 |
| 2868_rte                 |     3.732 |     0.304 |    12.29 |
| 2869_pegase              |     4.711 |     0.344 |    13.70 |
| 3012wp_k                 |     3.928 |     0.316 |    12.42 |
| 3022_goc                 |     4.858 |     0.525 |     9.26 |
| 3120sp_k                 |     4.096 |     0.371 |    11.03 |
| 3375wp_k                 |     5.250 |     0.453 |    11.59 |
| 3970_goc                 |     9.686 |     0.667 |    14.53 |
| 4020_goc                 |    10.059 |     0.770 |    13.06 |
| 4601_goc                 |    12.494 |     0.904 |    13.83 |
| 4619_goc                 |    14.324 |     1.052 |    13.61 |
| 4661_sdet                |    11.810 |     0.881 |    13.40 |
| 4837_goc                 |    14.117 |     1.032 |    13.68 |
| 4917_goc                 |    13.878 |     1.059 |    13.10 |
| 6468_rte                 |    25.386 |     2.005 |    12.66 |
| 6470_rte                 |    25.223 |     1.751 |    14.41 |
| 6495_rte                 |    26.308 |     2.035 |    12.93 |
| 6515_rte                 |    26.777 |     1.826 |    14.67 |

Those timings were obtained with `@time calc_basic_ptdf_matrix(data)`. It's not the most robust benchmark, but there's a clear speedup. Machine specs (via Julia's `versioninfo`):
```
Julia Version 1.8.1
Commit afb6c60d69a (2022-09-06 15:09 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × 12th Gen Intel(R) Core(TM) i7-1260P
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, goldmont)
  Threads: 1 on 16 virtual cores
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ PowerModels.jl Change Log
 
 ### Staged
 - Improved network data model documentation
+- Faster PTDF matrix computation (#849)
 
 ### v0.19.7
 - Improve linear algebra test robustness (#827)
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ The primary developer is Carleton Coffrin (@ccoffrin) with support from the foll
 - Miles Lubin (@mlubin) MIT, Julia/JuMP advise
 - Yeesian Ng (@yeesian) MIT, Documenter.jl setup
 - Kaarthik Sundar (@kaarthiksundar) LANL, OBBT utility
+- Mathieu Tanneau (@mtanneau) Georgia Tech, PTDF matrix computation
 - Byron Tasseff (@tasseff) LANL, multi-infrastructure updates
 
 
--- a/src/core/admittance_matrix.jl
+++ b/src/core/admittance_matrix.jl
@@ -136,7 +136,18 @@ end
 Base.show(io::IO, x::AdmittanceMatrixInverse{<:Number}) = print(io, "AdmittanceMatrixInverse($(length(x.idx_to_bus)) buses, $(length(x.matrix)) entries)")
 
 
-"note, data should be a PowerModels network data model; only supports networks with exactly one refrence bus"
+"""
+    calc_susceptance_matrix_inv(data)
+
+Compute the inverse of the network's susceptance matrix.
+
+Note: `data`` should be a PowerModels network data model; only supports networks with exactly one refrence bus.
+
+While the susceptance matrix is sparse, its inverse it typically quite dense.
+This implementation first computes a sparse factorization, then recovers the (dense)
+    matrix inverse via backward substitution. This is more efficient
+    than directly computing a dense inverse with `LinearAlgebra.inv`.
+"""
 function calc_susceptance_matrix_inv(data::Dict{String,<:Any})
     #TODO check single connected component
     sm = calc_susceptance_matrix(data)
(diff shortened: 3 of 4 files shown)


ccoffrin:
Nice!

These `basic` functions were designed to be more of a tutorial in nature for teaching the basics of power systems. Would it still get the performance boost if you move this implementation approach into the general `calc_susceptance_matrix_inv` function?  If so, this may be a more suitable location.

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/admittance_matrix.jl#L139-L166

Side note, for very large problems (say 10k or more buses) I find it is even faster to compute the PTDF cuts row-by-row, so you never have to compute the full inverse,

https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/util/flow_limit_cuts.jl#L87-L194

mtanneau:
>  Would it still get the performance boost if you move this implementation approach into the general calc_susceptance_matrix_inv function?

Let me check

> I find it is even faster to compute the PTDF cuts row-by-row

I have another piece of code to compute PTDF rows lazily, which itself is ~100x faster than what I wrote here (but it needs a cached factorization, so it's less user-friendly).

mtanneau:
> Would it still get the performance boost if you move this implementation approach into the general calc_susceptance_matrix_inv function?

Partly. Just replacing the two-level loop
https://github.com/lanl-ansi/PowerModels.jl/blob/cf37affe376358592d83851a60bc0270b79e0a4e/src/core/data_basic.jl#L343-L352
with a matrix-matrix product
```julia
B = calc_basic_branch_susceptance_matrix(data)
ptdf = B * b_inv
```
gives a ~5x speedup, which brings it within a factor 2-3 of my current code.

If I also replace the `calc_susceptance_matrix_inv` function to use a more efficient factorization (instead of dense inverse computation), then the computing times become similar.

ccoffrin:
I like your proposal for updating `calc_susceptance_matrix_inv` and then doing,
```
B = calc_basic_branch_susceptance_matrix(data)
ptdf = B * b_inv
```
I think it will keep this example simple and give the improved performance you are looking for in a broader collection of use cases.

ccoffrin:
Tests are passing so I am going to merge this.  House keeping before merge,
- Add a note the changelog
- Add your name to the README file as a contributor
- (optional) Add 1-2 sentences of documentation to the `calc_susceptance_matrix_inv` function on what this computational approach is and why in this context it is better than the Julia default / naive approach I had previously used.

PR #861, opened 2023-05-22
Reduce tolerance for change in Julia v1.9

odow:
PowerModels is failing tests on Julia 1.9 due to a tolerance issue:

image

### DIFF
--- a/test/am.jl
+++ b/test/am.jl
@@ -42,7 +42,7 @@ end
 
         @test isa(sm, AdmittanceMatrix{Float64})
         @test SparseArrays.nnz(sm.matrix) == 17
-        @test isapprox(LinearAlgebra.det(sm.matrix), 0.0, atol=1e-6)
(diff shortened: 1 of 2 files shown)


ccoffrin:
I and fine with this change. Any idea why it is still failing on windows Julia v1 for this one power flow case?

odow:
> Any idea why it is still failing on windows Julia v1 for this one power flow case?

Nope. Iteration limit, so I guess some numerical issue? It seems like lots of nearby tests are commented out with numerical difficulties.

ccoffrin:
@odow we can add this to the list of mysterious changes in results with v1.9, https://github.com/lanl-ansi/PowerModelsAnnex.jl/actions/runs/5063454342/jobs/9090086852

These differences do seem to be outside the bounds of machine precision issues.  I was not able to reproduce the issue in my intel based MacBook pro.

Issue #868, opened 2023-08-08
Piecewise-linear cost objective calculation


opfsolver:
I'm struggling to understand the [piecewise-linear cost computation](https://github.com/lanl-ansi/PowerModels.jl/blob/e9712cd1a162b546f9bff3e6d3f5d0816c22a3d5/src/core/objective.jl#L480-L511) of the objective. Apologies for the long text: I thought writing it out might answer my question, but unfortunately it didn't. Hopefully writing my thoughts down as clearly as possible will allow you to spot the errors in my understanding.

## Input data

Maybe the problem begins early, with my understanding of the cost data (`gen["cost"]`). IIUC `gen["cost"]` comes as a vector of real values, representing the piecewise linear blocks:

`gen["cost"] = [mw1, cost1, mw2, cost2, ...]`

where `mw1` is the total power _of the block_, and `cost1` is the marginal cost of that block. This means that if we want to produce `mw1 + mw2` power, we generate `mw1` amount using the marginal cost `cost1`, and `mw2` using the marginal cost `cost2`, with the total cost being `mw1*cost1 + mw2*cost2`.

The reason I think this is because this is the format used in the GOC2 competition, and the `PowerModelsSecurityConstrained.jl` code [uses](https://github.com/lanl-ansi/PowerModelsSecurityConstrained.jl/blob/1d72c2caab6d31230e8d663be5afc97fe6fbfd51/src/prob/opf.jl#L381) the `objective_variable_pg_cost` function.

## `calc_pwl_points`

My reading is that this function wants to remove the inactive blocks outside of the `pmin` and `pmax` range. It first constructs a vector of `points`, which is simply the same block cost data, stored as a list of named tuples.

Comment: I _suspect_ that `points` is meant to be something else here: the cumulative values of `mwN` and `costN` above, representing the points where different piecewise-linear segments meet.

With the cumulative interpretation, the next steps 
https://github.com/lanl-ansi/PowerModels.jl/blob/e9712cd1a162b546f9bff3e6d3f5d0816c22a3d5/src/core/objective.jl#L412-L430 make sense:

The `first_active` represents the index of the block that is covering the `pmin`, and `last_active` represents the index of the block covering `pmax`.

[...]

PR #869, opened 2023-08-09
Add links to documentation

opfsolver:
Adds references from https://github.com/lanl-ansi/PowerModels.jl/issues/868


ccoffrin:
Thanks for the proposal. I have addressed this point via this commit, https://github.com/lanl-ansi/PowerModels.jl/commit/27b564e366f7949d72064e15c7fbb6fcef0f98bc

PR #875, opened 2023-08-11
Add basic support for PST in DC OPF models

### LINKED ISSUE #543: Preliminary implementation of a controllable phase shifter

ccoffrin:
closes #543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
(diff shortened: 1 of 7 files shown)


ccoffrin:
Adding a new formulation `_solve_opf_pst` which only addresses pst optimization instead of both oltc and pst, `_solve_opf_oltc_pst`. This enables a clean implementation of a DCP and ACP formulations without having to consider the impact on oltc on the DCP approximation.

PR #879, opened 2023-08-13
Add test for JuMP Direct Mode

### LINKED ISSUE #275: JuMP Direct Mode
See if there is any need to consider using JuMP direct mode and/or expose this as parameter.

ccoffrin:
@odow does this look suitable for a direct mode test?

Closes #275

### DIFF
--- a/test/model.jl
+++ b/test/model.jl
@@ -21,6 +21,15 @@
         @test pm.model[:my_var] == x
         @test m[:my_var] == x
     end
+
+    @testset "run with user provided JuMP model in direct mode" begin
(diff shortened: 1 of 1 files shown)


odow (approved):

odow:
Yip. I don't think you need to do anything internally.

odow:
Maybe people haven't complained because they've just been doing this already.

ccoffrin:
Yes, indeed!  `jump_model` is not a well known feature but I can imagine such advanced users may have found it.

PR #880, opened 2023-08-17
Add DC alias for constraint_ohms_y_oltc_pst

tso-martin:
Dear @ccoffrin,

Thank you for integrating the constraint_ohms_y_pst constraints into the master. I added alias functions for the DCP formulations of `constraint_ohms_y_oltc_pst_to` and `constraint_ohms_y_oltc_pst_from` functions which refer to the newly added `constraint_ohms_y_pst_to` and `constraint_ohms_y_pst_from` in dcp.jl. 
In this way, it suffices to call constraint_ohms_y_oltc_pst from constraint_template.jl for both AC and DC formulations. The DC formulation obviously lacks the possibility to change the voltages by adjusting the OLTC (tm). But this should be clear if the DC models are used.

Best regards,
Martin

### DIFF
--- a/src/form/dcp.jl
+++ b/src/form/dcp.jl
@@ -155,7 +155,15 @@ function constraint_ohms_y_pst_to(pm::AbstractDCPModel, n::Int, f_bus, t_bus, f_
     JuMP.@constraint(pm.model, p_to == -b*(va_to - va_fr + ta))
 end
 
+""" Alias for constraint_ohms_y_pst_from"""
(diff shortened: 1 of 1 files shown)


ccoffrin:
Hi Martin, for now I would like to keep these two formulations (pst and oltc+pst) as separate things.  I do think it is possible and reasonable to include the impacts of oltc on the DC formulation, but this is not supported for now. For comparisons across AC and DC models that are supported now only pst is supported.

tso-martin:
Hi Carlton, ok, that's fine.

PR #886, opened 2023-09-10
Set global _LOGGER in __init__

### LINKED ISSUE #830: "IOError: stream is closed or unusable" when using parse_file() in Pluto.jl
Hello all,

I have encountered a new bug on my machine over the past few months when attempting to use `parse_file()` within a [Pluto.jl notebook.](https://www.juliapackages.com/p/pluto). This is one of those tricky issues where it's not clear which package is causing the bug, so I apologize in advance for the confusion. I looked into the source of both packages to try to seek out the root cause, but unfortunately I couldn't figure it out on my own. Please let me know if you agree that I should also open an issue with [Pluto](https://duckduckgo.com/?q=pltuo+github&). 

### Info on my machine:
- Julia Version: 1.7.3 (2022-05-06) Fedora 36 Build
- OS: Fedora 36
- CPU: Ryzen 4750U

### Step 1: setup environment to reproduce:
For convenience, you could skip this step and clone my repository I made to demonstrate this issue ([available here](https://github.com/samtalki/PlutoPowerModelsTest)).

Alternatively, make a fresh Julia environment >=1.7 and load your favorite test case into the root directory. Run these commands in the REPL:
```julia
pkg> generate PlutoPowerModelsTest
shell> cd PlutoPowerModelsTest
pkg> activate .
pkg> add Pluto
julia> import Pluto
julia> Pluto.run()
```

### Step 2: run `parse_file()` in a Pluto.jl notebook
Within the Pluto landing page, create a new notebook. Drop your `case14.m` or other file into the same directory as the notebook and then run the following commands within the notebook to reproduce the error. 
```julia
using PowerModels
case14 = parse_file("case14.m")
```
You should receive the following error:
image

ccoffrin:
from @odow hopefully fixes #830

### DIFF
--- a/src/PowerModels.jl
+++ b/src/PowerModels.jl
@@ -13,12 +13,11 @@ import InfrastructureModels
 import InfrastructureModels: optimize_model!, @im_fields, nw_id_default
 const _IM = InfrastructureModels
 
-# Create our module level logger (this will get precompiled)
-const _LOGGER = Memento.getlogger(@__MODULE__)
-
-# Register the module level logger at runtime so that folks can access the logger via `getlogger(PowerModels)`
-# NOTE: If this line is not included then the precompiled `PowerModels._LOGGER` won't be registered at runtime.
-__init__() = Memento.register(_LOGGER)
+function __init__()
+    global _LOGGER = Memento.getlogger(@__MODULE__)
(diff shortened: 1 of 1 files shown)


ccoffrin:
closing as did not fix.

PR #887, opened 2023-09-11
Remove Memento.jl and use Base Julia logging

ChrisRackauckas:
It doesn't seem to make a difference, other than fixes https://github.com/lanl-ansi/PowerModels.jl/issues/830


ChrisRackauckas:
Also fixes the benchmarks! https://github.com/SciML/SciMLBenchmarks.jl/pull/706

ccoffrin:
Closing as tests do not pass but good to know this is a suitable workaround.

PR #898, opened 2024-01-01
Updates for SCS v2.0

ccoffrin:
@odow, looks like SCS v2.0 changes have resolved the issue were CI results did not match my local system.  Yey!


odow:
I don't know why or how, but cool cool. We removed the shenanigans of how the MKL stuff was built.

PR #911, opened 2024-04-17
Fix docstring of DCPPowerModel

### LINKED ISSUE #910: Documentation: DCPPowerModel description text
Formula in the description of DC model appears to mistyped:
[link](https://lanl-ansi.github.io/PowerModels.jl/stable/formulation-details/#PowerModels.ACRPowerModel)
`br_b = -br_x / (br_x^2 + br_x^2)`

Instead I think it should be:
`br_b = -br_x / (br_x^2 + br_r^2)`

Can we fix this?

odow:
Closes #910 

I don't have the domain knowledge to know whether this is correct.

### DIFF
--- a/src/core/types.jl
+++ b/src/core/types.jl
@@ -156,7 +156,7 @@ abstract type AbstractDCPModel <: AbstractActivePowerModel end
 Linearized 'DC' power flow Model with polar voltage variables.
 
 This model is a basic linear active-power-only approximation, which uses branch susceptance values
-`br_b = -br_x / (br_x^2 + br_r^2)` for determining the network phase angles.  Furthermore, transformer
+`br_b = -br_x / (br_r^2 + br_x^2)` for determining the network phase angles.  Furthermore, transformer
 parameters such as tap ratios and phase shifts are not considered as part of this model.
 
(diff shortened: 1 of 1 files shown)


src/core/types.jl
```diff
@@ -156,7 +156,7 @@ abstract type AbstractDCPModel <: AbstractActivePowerModel end
 Linearized 'DC' power flow Model with polar voltage variables.
 
 This model is a basic linear active-power-only approximation, which uses branch susceptance values
-`br_b = -br_x / (br_x^2 + br_x^2)` for determining the network phase angles.  Furthermore, transformer
+`br_b = -br_x / (br_x^2 + br_r^2)` for determining the network phase angles.  Furthermore, transformer
```
ccoffrin, line 159: This is a correct fix but the convention is that r comes before x, so `(br_r^2 + br_x^2)` would be more standard.

the reason is the complex number `Z = r + x im`, then `|Z|^2 = r^2 + x^2`

odow, line 159:
```suggestion
`br_b = -br_x / (br_r^2 + br_x^2)` for determining the network phase angles.  Furthermore, transformer
```

ccoffrin:
Thanks!

Issue #913, opened 2024-06-11
Performance of make_basic_network


mtanneau:
The `make_basic_network` function is slow for large networks.
The timings below were obtained with `@time make_basic_network(data)` where `data` was loaded with the PGLib package.

| Case | #buses | time (s) | memory (MiB) |
|:--|--:|--:|--:|
| `1354_pegase` | 1354 | 0.21 | 28.5 |
| `2869_pegase` | 2869 | 0.62 | 66.8 |
| `9241_pegase` | 9241 | 2.50 | 208.4 |
| `13659_pegase` | 13659 | 5.72 | 303.1 |
| `30000_goc` | 30000 | 15.81 | 467.8 |

I profiled it on the `9241_pegase` case:
![prof_old](https://github.com/lanl-ansi/PowerModels.jl/assets/9593025/9b595704-f031-4f1b-97ef-be101139ee1f)

which shows that, besides an initial `deepcopy` which we can't do much about, a substantial amount of time is spent identifying connected components.
Note that the very high stacktrace with `_cc_dfs` corresponds to recursive calls to a DFS implementation [here](https://github.com/lanl-ansi/PowerModels.jl/blob/8e346b48c187b2a53907dc031603482f2e6a2017/src/core/data.jl#L2547-L2560).

PR coming

mtanneau:
@ccoffrin I have opened two PRs to address the initial deepcopy (#915) and the identification of connected components (#914).

The 3rd room for performance improvement is within [`_renumber_components`](https://github.com/lanl-ansi/PowerModels.jl/blob/8e346b48c187b2a53907dc031603482f2e6a2017/src/core/data_basic.jl#L151-L163), which contains a lot of `deepcopy`s itself.
Namely, all individual components are deepcopied before being re-numbered.

I believe this is not necessary, since we deepcopy the initial data dictionary, and everything afterwards modifies the data dict in-place. As far as I can see, `_renumber_components` is only called inside `make_basic_network`, so eliminating the deepcopy below would save further time and memory
https://github.com/lanl-ansi/PowerModels.jl/blob/8e346b48c187b2a53907dc031603482f2e6a2017/src/core/data_basic.jl#L156-L160

ccoffrin:
Has this been addressed by https://github.com/lanl-ansi/PowerModels.jl/pull/919?

mtanneau:
Yes!

For completeness, I re-ran the timings and profiles; results are below.

___

[...]

PR #915, opened 2024-06-12
Add in-place conversion to basic data format

mtanneau:
Motivated by #913 

This PR introduces an in-place `make_basic_data!`, which has the same functionality but does not create a `deepcopy`.
The goal is to avoid the (time and memory) cost of deepcopy-ing the initial data dictionary when the user does not want to keep it.

A typical use-case would be workflows that look like `data = make_basic_network(parse_file(...))`.
This call with load the raw data dictionary with `parse_file`, then deep-copy it, convert the copy to basic format, and discard the original raw data.

Timings of `make_basic_network` after the changes: times are improved, and memory consumption is reduced.
Note: those are just the ouptut of `@time`; I did run `GC.gc()` before to reduce garbage collection, but it's not a full-fledge `@benchmark`

| Case | #buses | old time (s) | in-place time (s) | old memory (MiB) | in-place memory (MiB) |
|:--|--:|--:|--:|--:|--:|
| `1354_pegase` | 1354 | 0.21 | 0.13 | 29.4 | 20.4 |
| `2869_pegase` | 2869 | 0.60 | 0.40 | 66.7 | 48.2 |
| `9241_pegase` | 9241 | 2.50 | 1.87 | 208.3 | 145.9 |
| `13659_pegase` | 13659 | 5.44 | 4.67 | 313.0 | 227.4 |

and corresponding profile of `make_basic_network!(data)` where `data = pglib("9241_pegase")`:
prof_inplace

Note that the above profile does not include the improvements in #914. The main (and expected) outcome is that the time spent deep-copying at the beginning is gone.

### DIFF
--- a/test/data-basic.jl
+++ b/test/data-basic.jl
@@ -35,6 +35,35 @@
         @test isapprox(result["objective"], 16551.7; atol=1e0)
     end
 
+    @testset "In-place make_basic_network!" begin
+        data = PowerModels.parse_file("../test/data/matpower/case7_tplgy.m")
+        data_ = deepcopy(data)
+        
+        # Test #1: `make_basic_network` is non-destructive
+        data_basic = PowerModels.make_basic_network(data)
+        @test data == data_        # input data not modified
+        @test data_basic !== data  # we created a new dict
+
+        # Test #2: `make_basic_network!` is destructive and works in-place
+        data_basic = PowerModels.make_basic_network!(data)
+        @test data_basic === data  # in-place modification
+
+        # Test #3: double-check the output of `make_basic_network!`
+        #   This may be redundant with the above tests, unless `make_basic_network` does not
+        #   call the in-place version under the hood
+        @test length(data_basic["bus"]) == 4
+        @test length(data_basic["load"]) == 2
+        @test length(data_basic["shunt"]) == 2
+        @test length(data_basic["gen"]) == 1
+        @test length(data_basic["branch"]) == 2
+        @test length(data_basic["dcline"]) == 0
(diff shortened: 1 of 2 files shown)


ccoffrin:
Shall we add a test for `make_basic_network!` and `make_basic_network` that confirms that one is destructive and the other is not?

mtanneau:
I can do that. I'm currently in a plane, will push in a couple days

ccoffrin:
After that I'll be good to merge.  Feel free to add your name in the list of the readme if you like.

test/data-basic.jl
```diff
@@ -35,6 +35,35 @@
...
     end
 
+    @testset "In-place make_basic_network!" begin
+        data = PowerModels.parse_file("../test/data/matpower/case7_tplgy.m")
+        data_ = deepcopy(data)
+        
+        # Test #1: `make_basic_network` is non-destructive
+        data_basic = PowerModels.make_basic_network(data)
+        @test data == data_        # input data not modified
+        @test data_basic !== data  # we created a new dict
```
mtanneau, line 45: These ones check that we did not modify the input data, and that we returned a new `Dict` object

test/data-basic.jl
```diff
@@ -35,6 +35,35 @@
...
+
+        # Test #2: `make_basic_network!` is destructive and works in-place
+        data_basic = PowerModels.make_basic_network!(data)
+        @test data_copy === data  # in-place modification
```
mtanneau, line 49: Here I only check that the returned dictionary is exactly the one that was passed as input.
Note the use of triple equal `===`

PR #916, opened 2024-06-12
No deepcopy when renumbering components

mtanneau:
Motivated by #913. 

This PR removes an internal `deepcopy` while re-indexing components in `_renumber_components`, thus reducing time and memory allocations.

| Case | #buses | old time (s) | in-place time (s) | old memory (MiB) | in-place memory (MiB) |
|:--|--:|--:|--:|--:|--:|
| `1354_pegase` | 1354 | 0.21 | 0.15 | 29.4 | 20.0 |
| `2869_pegase` | 2869 | 0.60 | 0.2 | 66.7 | 45.7 |
| `9241_pegase` | 9241 | 2.50 | 1.91 | 208.3 | 136.5 |
| `13659_pegase` | 13659 | 5.44 | 4.63 | 313.0 | 215.6 |

and corresponding profile of `make_basic_network(data)` where `data = pglib("9241_pegase")`

prof_no_internal_deepcopy

### DIFF
--- a/src/core/data_basic.jl
+++ b/src/core/data_basic.jl
@@ -110,7 +110,7 @@ function make_basic_network(data::Dict{String,<:Any})
     # re-number non-bus component ids
     for comp_key in keys(pm_component_status)
         if comp_key != "bus"
-            data[comp_key] = _renumber_components(data[comp_key])
+            data[comp_key] = _renumber_components!(data[comp_key])
         end
     end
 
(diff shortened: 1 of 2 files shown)


ccoffrin (commented):
For this one I cannot immediately confirm the change is ok.

The function `_renumber_components` does not have a bang, so it should be non-destructive for the input arguments.  By changing the value of `comp["index"]` without copying it, some of the data in `comp_dict` has been changed.

So I am suspecting this function should be renamed `_renumber_components!` and then it should be confirmed that all existing uses are still OK when this function is destructive.

mtanneau:
That is true 🤔

I believe an in-place, destructive version would serve us better indeed. No need for a new dictionary, we can update the existing one in-place, which would also make things more explicit to the user.
Currently, we do not modify the key -> value mapping of the input dictionary directly, we modify `comp_dict[i]["index"]`.

> So I am suspecting this function should be renamed _renumber_components! and then it should be confirmed that all existing uses are still OK when this function is destructive.

I did a repo-wide search of `_renumber_components`, and I only found it used once, as part of `make_basic_network` (which already created a deepcopy of the original data dictionary). I believe changing it to a destructive version `_renumber_components!` would not affect code within PowerModels.
To avoid any downstream issue, I can do the following:
* create an in-place, destructive version `_renumber_components!` 
* update the existing `_renumber_components` to be something like `_renumber_components!(deepcopy(input_data))`

If you want to get rid of the non-destructive version, it can be deprecated and removed in a later version.

ccoffrin:
> I did a repo-wide search of _renumber_components, and I only found it used once, as part of make_basic_network (which already created a deepcopy of the original data dictionary). I believe changing it to a destructive version _renumber_components! would not affect code within PowerModels.

If that is the case, I am fine to change this to the destructive version along with the function name change.  I don't mind removing the original.  The change is "non-breaking" to external users because this function is indicated as internal use only and not exported.

Issue #912, opened 2024-05-28
PSSE Parser: incorrect impedance when NOMVX is 0.0 


frederikgeth:
Ran into this bug with @hei06j while parsing 3-winding transformers from PSSE.

Per the PSSE Manual on NOMV1, NOMV2, NOMV3: 
> The nominal (rated) winding one voltage in kV, or zero to indicate that nominal winding one voltage is to be taken as the base voltage of bus "I". NOMV1 is used only in converting magnetizing data between per unit admittance values and physical units when CM is 2. NOMV1 = 0.0 by default.

If that happens, the transformer impedances get accidentally set to 0 [here](https://github.com/lanl-ansi/PowerModels.jl/blob/8e346b48c187b2a53907dc031603482f2e6a2017/src/io/psse.jl#L437)
```julia
br_r *= (transformer["NOMV1"]^2 / _get_bus_value(transformer["I"], "base_kv", pm_data)^2) * (pm_data["baseMVA"] / transformer["SBASE1-2"])
br_x *= (transformer["NOMV1"]^2 / _get_bus_value(transformer["I"], "base_kv", pm_data)^2) * (pm_data["baseMVA"] / transformer["SBASE1-2"])
```
This should be replaced with something like
```julia
if transformer["NOMV1"] == 0.0
  br_r *= (pm_data["baseMVA"] / transformer["SBASE1-2"])
  br_x *= (pm_data["baseMVA"] / transformer["SBASE1-2"])
else
  br_r *= (transformer["NOMV1"]^2 / _get_bus_value(transformer["I"], "base_kv", pm_data)^2) * (pm_data["baseMVA"] / transformer["SBASE1-2"])
  br_x *= (transformer["NOMV1"]^2 / _get_bus_value(transformer["I"], "base_kv", pm_data)^2) * (pm_data["baseMVA"] / transformer["SBASE1-2"])
end
```

frederikgeth:
Just saw Jay's issue and I think it is the same underlying issue, https://github.com/lanl-ansi/PowerModels.jl/issues/888, which ran into this problem for branches, not (3-winding) transformers

Note that with r and x both effectively becoming zero none of the bus injection model type formulations will work, and instead converge to an infeasible point.

ccoffrin:
Can you adapt one of the test network datasets to exhibit the issue so that we can make a test for this?

frederikgeth:
[...]

Issue #888, opened 2023-09-18
Not able to solve pti model


jay-dave:
I am able to solve power flow on attached model using PSSE but not using PowerModels (shows invalid model). Below is the code to run the attached example model. Can someone help me to figure out the issue?
------------------------------------------------------------------------------------------------------------------------
using PowerModels
import InfrastructureModels
import Ipopt
import JuMP
import JSON

if !isdefined(JuMP, :num_nonlinear_constraints)
    num_nonlinear_constraints = JuMP.num_nl_constraints
else
    num_nonlinear_constraints = JuMP.num_nonlinear_constraints
end
nlp_solver = JuMP.optimizer_with_attributes(Ipopt.Optimizer, "tol"=>1e-6, "print_level"=>2)

Pglib_file = "yourpath/example.raw"
#setting generator scheduled voltage as bus voltage
source_data = parse_file(Pglib_file)
for (g,gen) in source_data["gen"]
    genbus = gen["gen_bus"]
    source_data["bus"]["$genbus"]["vm"] = gen["vg"]
end
Pglib_solution = PowerModels.run_pf(source_data, ACPPowerModel, nlp_solver, setting = Dict("output" => Dict("branch_flows" => true, "duals" => true)))

[PTIcase-main.zip](https://github.com/lanl-ansi/PowerModels.jl/files/12648039/PTIcase-main.zip)

jay-dave:
I think this is somewhat related to issue #737. when I changed the CZ and CW to 1, the branch reactance takes nonzero value and it works. However, it was not working for CZ = 3 and CW = 2.

On a closer look, it seems that pti assumes transformer base kV = Bus base kV when NOM1, NOM2 (or NOM3 for three winding transformers) =0. However, it is taken as zero in data processing in PM. Hence, the branch reactance becomes zero during pu conversion and causes model to be invalid.

hei06j:
This issue is mostly addressed in pull request #917, but the case in this issue is still problematic due to transformer sbase values. With the PR in place, I could solve this example with Sbase>70, but Sbase=60 still does not solve. Note that one of the unit tests as part of this PR is the example in this issue with Sbase=100.

PR #917, opened 2024-06-16
Update psse parser for 3winding transformers

hei06j:
This PR is to address the issue #912, which also partly addresses the issue #888. The issue #888 still stands, but with this PR, it solves by changing the transformer sbase to >70 (it is 60 now and not solving).

PSSE three-winding transformer supports 
- zero NOMVx values, which in this case, the transformer x-th winding nominal voltage is the same as the bus basekV it is connected to,
- cw = 1 (for off-nominal turns ratio in pu of winding bus base voltage)
- cw = 2 (for winding voltage in kv)
- cw = 3 (for off-nominal turns ratio in pu of nominal winding voltage, NOMV1, NOMV2 and NOMV3.)

which are now supported in the PSSE parser as part of this PR.


src/io/psse.jl
```diff
@@ -601,9 +632,12 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
-                        if transformer["CW"] == 3  # "for off-nominal turns ratio in pu of nominal winding voltage, NOMV1, NOMV2 and NOMV3."
-                            sub_data["tap"] *= transformer["NOMV$m"]
+                        if transformer["CW"] == 2
+                            sub_data["tap"] /= _get_bus_value(bus_id, "base_kv", pm_data)
+                        elseif transformer["CW"] == 3  # "for off-nominal turns ratio in pu of nominal winding voltage, NOMV1, NOMV2 and NOMV3."
+                            if !iszero(transformer["NOMV$m"])
+                                sub_data["tap"] *= transformer["NOMV$m"] / _get_bus_value(bus_id, "base_kv", pm_data)
+                            end
                         end
                     end
```
ccoffrin, line 642: Could you elaborate about what the correct semantics of this should be?

The way the code is organized seems risky to me.  For example, what happens if `transformer["CW"] == 3` and `iszero(transformer["NOMV$m"]) == true`

hei06j: `iszero(transformer["NOMV$m"]) == true` means that the transformer winding `m` nominal voltage is the same as the bus base voltage it is connected to, i.e. `_get_bus_value(bus_id, "base_kv", pm_data)`. So in the case of `transformer["CW"] == 3`, the tap value won't change. For when `transformer["CW"] == 2`, the winding voltage is in kv and does not need to account for `transformer["NOMV$m"]`.

ccoffrin: Given the complexity of this sepecific section I suggest we re-write the code int he form,

```
if transformer["CW"] == 1
  ...
elseif transformer["CW"] == 2
  ...
elseif transformer["CW"] == 3
  ...
else
  error - unsupported value for "CW" on transformer ....
end
```

hei06j: Recommendation implemented. :)

src/io/psse.jl
```diff
@@ -544,13 +562,29 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
 
                 # Unit Transformations
                 if transformer["CZ"] != 1  # NOT "for resistance and reactance in pu on system MVA base and winding voltage base"
-                    br_r12 *= (transformer["NOMV1"] / _get_bus_value(bus_id1, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE1-2"])
-                    br_r23 *= (transformer["NOMV2"] / _get_bus_value(bus_id2, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE2-3"])
-                    br_r31 *= (transformer["NOMV3"] / _get_bus_value(bus_id3, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE3-1"])
+                    if transformer["NOMV1"] == 0.0
```
ccoffrin, line 565: same note as above.

src/io/psse.jl
```diff
@@ -544,13 +562,29 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+                        br_x12 *= (pm_data["baseMVA"] / transformer["SBASE1-2"])
+                    else
+                        br_r12 *= (transformer["NOMV1"] / _get_bus_value(bus_id1, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE1-2"])
+                        br_x12 *= (transformer["NOMV1"] / _get_bus_value(bus_id1, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE1-2"])
+                    end
 
-                    br_x12 *= (transformer["NOMV1"] / _get_bus_value(bus_id1, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE1-2"])
-                    br_x23 *= (transformer["NOMV2"] / _get_bus_value(bus_id2, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE2-3"])
-                    br_x31 *= (transformer["NOMV3"] / _get_bus_value(bus_id3, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE3-1"])
+                    if transformer["NOMV2"] == 0.0
```
ccoffrin, line 573: same note as above.

src/io/psse.jl
```diff
@@ -544,13 +562,29 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+                        br_r23 *= (pm_data["baseMVA"] / transformer["SBASE2-3"])
+                        br_x23 *= (pm_data["baseMVA"] / transformer["SBASE2-3"])
+                    else
+                        br_r23 *= (transformer["NOMV2"] / _get_bus_value(bus_id2, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE2-3"])
+                        br_x23 *= (transformer["NOMV2"] / _get_bus_value(bus_id2, "base_kv", pm_data))^2 * (pm_data["baseMVA"] / transformer["SBASE2-3"])
+                    end
+                    
+                    if transformer["NOMV3"] == 0.0
```
ccoffrin, line 581: same note as above

ccoffrin (commented):
Ok, I think this is very close to ready. I do have one last question about float equality testing to resolve.

ccoffrin:
Looks good.  Thanks for the contribution!

PR #920, opened 2024-06-24
PTI Parser Quote Counter Fix

aniruddh-mishra:
Found a small bug that prevents the current PTI parser from throwing an error at mismatched quotations.

The "'" represents a string, while the count function expects to compare i with a character. This means that the current code always counted 0 quotes in each line. This small fix should correct the error handling in this case.


ccoffrin:
closing as fixed and test were added here, https://github.com/lanl-ansi/PowerModels.jl/pull/919/commits/17148a18d9de2ac1bcab36e2b1a68786a1f4cb34

PR #933, opened 2024-10-30
Fix error in `calc_connected_components` if there are no active buses

Robbybp:
By just returning an empty set.


ccoffrin:
LGTM. Thanks!

PR #934, opened 2024-11-01
Fix PSSE raw import for gen_status at load buses

### LINKED ISSUE #932: Incorrect behaviour for PSSE active generators at load buses
Default PSSE behaviour is to ignore (remove) active generators at load buses (type 1). PowerModels does the opposite while parsing the raw. It changes the bus type to generator (type 2) instead. PowerModels should adapt the IO script to mimic native PSSE behaviour.

One relevant part of the PowerModels code is located in `_correct_bus_types!` function.

KSepetanc:
Fixes #932 .

PSSE ignores generators at load buses (bus type 1). It does not use this flag to determine if the generator is PV or PQ which is different from Matpower. Instead, it uses `Machine Control Mode` parameter found in panel `Machine Data Record -> Wind Data -> Control Mode` or in the `Machine` sheet.

As such fix needs to be implemented only for PSSE and there are no changes for Matpower. Thus, the only change is in `src\io\psse.jl`.

Consider adding `Memento` warning before merging.


KSepetanc:
@ccoffrin can you review this please?

ccoffrin:
Sorry for the delay @KSepetanc. I have no way of verifying these semantics of PSSE.  @jd-lara or @frederikgeth are you able to verify this behavior?

@KSepetanc, before merging we will need to add a test for this behavior so that we don't have a future regression.

frederikgeth:
not able to verify, sorry, but can confirm he actual status of a machine is determined both by its status flag and the bus type code.

KSepetanc:
@ccoffrin I can not find any network in the PowerModels test/data/pti that has generator placed at load bus. I am not very comfortable modifying those frankenstein networks (presumably `frankenstein_70.raw`)  as I do not know repercussions on other tests. I will need help with this.

test construct should be something like this (in `test/psse.jl` after transformer status tests):

```
    @testset "generator status" begin
        data_pti = PowerModels.parse_file("../test/data/pti/case5.raw")

        gen_off = Set(["8", "9"])
        for (i,gen) in data_pti["gen"]
            if i in gen_off
                @test gen["gen_status"] == 0
            else
                @test gen["gen_status"] == 1
            end
        end
    end
```

KSepetanc:
I have modified `frankenstein_70` at but `1005` to load bus. Test has been added.

Do you think we need to add memento log in `src/io/psse.jl line 214` for the case when bus type is 1 and gen["STAT"] is 1?

KSepetanc:
@ccoffrin do you have someone else how can check this? I would like to close this issue soon.

KSepetanc:
There is free trial version of PSSE with full computation support for up-to 50 buses. Can that help with the review?

src/io/psse.jl
```diff
@@ -211,7 +211,7 @@ function _psse2pm_generator!(pm_data::Dict, pti_data::Dict, import_all::Bool)
             sub_data = Dict{String,Any}()
 
             sub_data["gen_bus"] = pop!(gen, "I")
-            sub_data["gen_status"] = pop!(gen, "STAT")
+            sub_data["gen_status"] = _get_bus_value(sub_data["gen_bus"], "bus_type", pm_data) == 1 ? begin delete!(gen, "STAT"); 0 end : pop!(gen, "STAT") #	In PSSE active generators at bus type 1 are ignored. PQ or PV control mode is controlled by "Machine Control Mode" parameter (Machine Data Record -> Wind Data -> Control Mode)
```
odow, line 214:
```suggestion
            if _get_bus_value(sub_data["gen_bus"], "bus_type", pm_data) == 1
                # In PSSE active generators at bus type 1 are ignored. PQ or PV
                # control mode is controlled by "Machine Control Mode" parameter
                # (Machine Data Record -> Wind Data -> Control Mode)
                delete!(gen, "STAT")
                sub_data["gen_status"] = 0
            else
                sub_data["gen_status"] = pop!(gen, "STAT") 
            end
```

odow:
> PSSE ignores generators at load buses (bus type 1)

Is there any documentation that we can link to that confirms this is the expected behavior? Otherwise, @KSepetanc could you construct an example in PSSE that confirms the problem? (e.g., with screenshots?)

KSepetanc:
Image below shows no power going out of generator 1 at the shown bus. Bus type is seen in the text on the left (type 1 which is load bus). Generator is active and set to produce 10 MW and 10 MVA (but does not inject anything). I used psse 35.1. I also have v34 and the result is the same. I will accept proposed revision.

image

odow:
Thanks. I don't have the domain expertise to know if this is correct, but it should help @ccoffrin review without needing to do work on his side.

KSepetanc:
@ccoffrin, just a reminder, what is your feedback on this? Do you have additional requirements?

PR #939, opened 2025-01-24
Fix element type of Dict in pf.jl

### LINKED ISSUE #938: Using integer values in input data Dict leads to an InexactError in compute_ac_pf()
When trying to compute the ac power flow for my model, it didn't work as I had integer zeroes (0) instead of float zeroes (0.0) in my input dictionary.
Attached is the stacktrace and two very similar cases, where one works and one doesn't. The only difference is the 0/0.0. 

Stacktrace:

```
ERROR: InexactError: Int64(-0.07035222288586446)
Stacktrace:
  [1] Int64
    @ .\float.jl:994 [inlined]
  [2] convert
    @ .\number.jl:7 [inlined]
  [3] setindex!(h::Dict{String, Int64}, v0::Float64, key::String)
    @ Base .\dict.jl:355
  [4] _assign_qg!(sol_gens::Dict{String, Any}, bus_gens::Vector{Any}, qg_remaining::Float64)
    @ PowerModels C:\Users\kuhrmann\.julia\packages\PowerModels\6U4MZ\src\prob\pf.jl:506
  [5] compute_ac_pf(pf_data::PowerFlowData; kwargs::@Kwargs{})
    @ PowerModels C:\Users\kuhrmann\.julia\packages\PowerModels\6U4MZ\src\prob\pf.jl:348
  [6] compute_ac_pf
    @ C:\Users\kuhrmann\.julia\packages\PowerModels\6U4MZ\src\prob\pf.jl:293 [inlined]
  [7] #compute_ac_pf#1003
    @ C:\Users\kuhrmann\.julia\packages\PowerModels\6U4MZ\src\prob\pf.jl:282 [inlined]
  [8] compute_ac_pf
    @ C:\Users\kuhrmann\.julia\packages\PowerModels\6U4MZ\src\prob\pf.jl:276 [inlined]
  [9] solve_ac_PF(solver_type::String, network_data::Dict{String, Any})
```

See attached models as examples:

[broken_model.json](https://github.com/user-attachments/files/18508132/broken_model.json)
[working_model.json](https://github.com/user-attachments/files/18508133/working_model.json)

The only code used was `parse_file()` and `compute_ac_pf()`.

The only difference is that pg in the working models is 0.0, while in the broken model it is 0.
I think think that this should maybe be caught somewhere in `correct_network_data!()`.
Also doesn't happen if the model is in mixed units (i assume the 0 gets divided by the unit base, leading to a 0.0).

Thanks for your efforts! Love the project!

odow:
Closes #938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
-- nothing
+- Fix InexactError in `compute_ac_pf` (#939)
 
 ### v0.21.3
 - Fix no-buses bug in `calc_connected_components` (#933)
--- a/src/prob/pf.jl
+++ b/src/prob/pf.jl
@@ -312,7 +312,7 @@ function compute_ac_pf(pf_data::PowerFlowData; kwargs...)
             if bus["bus_type"] != 4
                 bus_idx = am.bus_to_idx[bus["index"]]
 
-                bus_assignment[i] = Dict(
+                bus_assignment[i] = Dict{String,Float64}(
                     "vm" => pf_data.vm_idx[bus_idx],
                     "va" => pf_data.va_idx[bus_idx]
                 )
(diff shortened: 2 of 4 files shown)


ccoffrin:
Change is good by me, but should we add a test to make sure that we don't have a regression in the future?

PR #940, opened 2025-01-27
Add support for OPFData dataset format

akuramshin:
Adding a parser for the [OPFData dataset](https://arxiv.org/pdf/2406.07234). This parser is largely based on the existing parser for matpower files.


odow:
It is possible to use a smaller test example that isn't half a million lines?

akuramshin:
> It is possible to use a smaller test example that isn't half a million lines?

Sorry about that! Replaced with a smaller case file. Let me know if you would prefer an even smaller file

PR #941, opened 2025-01-27
Fixed dictionary call syntax error in psse export

LKuhrmann:
There is a very simple syntax error in the pm2psse transformer export function.
(A dict was being accessed with brackets instead of square brackets).

I did not find this issue, but a former colleague didn't pull request/ issue tracker it (so I don't really know the impact of the bug, just that my former colleague found it and fixed it locally).

If this is too small of a fix for a pull request, you are welcome to add it to a different commit/pr.


odow (approved):
Looks correct. I don't know enough details to find a test.

PR #942, opened 2025-02-07
Assign names to constraints

alimaala2002:
We are implementing reformulation techniques and faced difficulties accessing certain constraints. To address this, we assigned names to them for easier retrieval.


odow:
I get the motivation for this, but it seems very brittle.

I think this really needs PowerModels to have a much richer data structure for the variables and constraints that it adds, or some features in your code to automatically identify the type of structure that you can exploit through decomposition.

Perhaps more information would bee useful to understand the motivation. What are you trying to decompose and how? And how would the name make this easier?

odow:
Hi @alimaala2002, any update on the motivation for this? Otherwise I think I will close this issue. We don't want the API of PowerModels to depend on the `String` name of constraints.

ccoffrin:
Thanks for your suggestion. We discussed this topic internally and agreed this would be a major change that we are not ready to make in PowerModels at this time. We do note that a few constraints (e.g., nodal power balance) are tracked to support reporting the [duals on those constraints](https://github.com/lanl-ansi/PowerModels.jl/blob/master/test/output.jl#L140-L236). Inconsistencies in the support for these currently supported constraints are bugs that can be fixed.

We suspect that your use case would be better served by one of the from-scratch JuMP model implementations that are available in PowerModelsAnnex, https://github.com/lanl-ansi/PowerModelsAnnex.jl/blob/master/src/model/opf.jl

PR #943, opened 2025-02-13
More accurate message when deactivating isolated components

Robbybp:
Was confused by this message so I updated it to more accurately reflect the logic.


ccoffrin:
Change looks good, can you add a note to the change log, then merge into master?

PR #944, opened 2025-02-13
Added check for positive resistances and reactances

alexe15:
Thank you so much for this great package! I use it a lot, its my absolute favorite tool to do power system computations. 

I've added a check for positive resistances and reactances in make_basic_network. Currently it issues a warning if they are not positive, but maybe it should even be an error. I found cases with negative resistances in the pglib repo, cf. https://github.com/power-grid-lib/pglib-opf/issues/45

### DIFF
--- a/src/core/data_basic.jl
+++ b/src/core/data_basic.jl
@@ -76,6 +76,17 @@ function make_basic_network!(data::Dict{String,<:Any})
         end
     end
 
+    # check whether all resistances and conductances are > 0
+    for (i,branch) in data["branch"]
+        if branch["br_r"] < 0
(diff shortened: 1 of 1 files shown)


ccoffrin:
Hi @alexe15, I am glad you find power models so useful.

Thank you for the proposal, but I am not inclined to accept this change. Non-positive resistances and reactances are quite comment in real-world network models (as you notice in the PGLib-OPF datasets).

Warning about the existence of such branches is also unrelated to the specific objectives of the "basic data" utilities,
https://lanl-ansi.github.io/PowerModels.jl/stable/basic-data-utilities/

PR #948, opened 2025-03-03
Bump actions/cache version to v4

mtanneau:
This hot fix should fix the deprecated action issue.
PR #947 should prevent future similar issues in the future


odow:
Closing in favor of https://github.com/lanl-ansi/PowerModels.jl/pull/949

PR #958, opened 2025-04-21
Fix parse_file to use JSON.parsefile

odow:
Closes https://github.com/lanl-ansi/PowerModels.jl/pull/955


KSepetanc:
I read that JSON3.jl is recommended for parsing json. It should be more performant than JSON.
What are your thoughts about this?

odow:
I'd prefer to keep the JSON.jl dependency. Parsing time is a very very minor part of any workflow.

ccoffrin:
I had a quick look. No major objections to the merged version. Appears to be nonbreaking.

PR #959, opened 2025-04-21
Fix spelling of resolve_switches and add deprecation for compatibility

### LINKED ISSUE #936: Spelling Fix
`resolve_swithces!` -> `resolve_switches!`

odow:
Closes #936


ccoffrin:
Fix looks good to me.  Worth a discussion if we should put this change into the next patch or breaking release.

PR #961, opened 2025-04-22
Clean up package imports

odow:


src/PowerModels.jl
```diff
@@ -1,17 +1,14 @@
 module PowerModels
 
-import LinearAlgebra, SparseArrays
-
+import InfrastructureModels
+import InfrastructureModels as _IM
```
ccoffrin, line 4: Wondering why we have both of these lines.  Isn't one sufficient?

src/core/admittance_matrix.jl
```diff
@@ -17,7 +17,7 @@ struct AdmittanceMatrix{T}
     matrix::SparseArrays.SparseMatrixCSC{T,Int}
 end
 
-Base.show(io::IO, x::AdmittanceMatrix{<:Number}) = print(io, "AdmittanceMatrix($(length(x.idx_to_bus)) buses, $(length(nonzeros(x.matrix))) entries)")
+Base.show(io::IO, x::AdmittanceMatrix{<:Number}) = print(io, "AdmittanceMatrix($(length(x.idx_to_bus)) buses, $(SparseArrays.nnz(x.matrix)) entries)")
```
ccoffrin, line 20: Is this change non-breaking?  Not obvious to me.

src/core/admittance_matrix.jl
```diff
@@ -162,14 +162,14 @@ function calc_susceptance_matrix_inv(data::Dict{String,<:Any})
...
     S[ref_idx, ref_idx] = 1.0
-    
-    F = LinearAlgebra.ldlt(Symmetric(S); check=false)
+
+    F = LinearAlgebra.ldlt(LinearAlgebra.Symmetric(S); check=false)
     if !LinearAlgebra.issuccess(F)
         Memento.error(_LOGGER, "Failed factorization in calc_susceptance_matrix_inv")
     end
-    M = F \ Matrix(1.0I, num_buses, num_buses)
+    M = F \ Matrix{Float64}(LinearAlgebra.I(num_buses))
```
ccoffrin, line 170: Would like to understand this change.

ccoffrin (commented):
A few points to discuss in this one.

src/core/admittance_matrix.jl
odow: They're equivalent. `SparseArrays.nonzeros` returns a `Vector` of the non-zeros. `SparseArrays.nnz` returns the integer count.

src/PowerModels.jl
odow: There was a test that accessed the full `InfrastructureModels` symbol:
https://github.com/lanl-ansi/PowerModels.jl/blob/95e6c2cb54e2fbe9b9c05c7e9e2618fe117e6d0b/test/matpower.jl#L203

Ideally, you wouldn't do this, and we could remove the first line, which would make it more obvious that the `_IM` symbol is private.

src/core/admittance_matrix.jl
odow: I tried `M = F \ LinearAlgebra.I(num_buses)`, but I got a weird method error.

We could have kept `M = F \ Matrix(1.0 * LinearAlgebra.I, num_buses, num_buses)`, but I prefer the `I(n)` constructor.

I dislike the non-LinearAlgebra prefixed `I` because it gets too easily confused with a bunch of other `I`, particularly when we have `I, J, V` for sparse matrices.

ccoffrin:
All good.

Issue #73, opened 2017-01-29
Speed up Tests


ccoffrin:
See if caching parsed versions of the model data can improve the speed of running package tests.

odow:
Closing as won't fix. The tests run in 5-6 minutes, which seems okay. And with caching of packages etc, the total run time is now much better than it was in 2017.

ccoffrin:
I am not sure I am ready to close this one yet.

The heart of the issue was, and I think still is, that there are a small-ish number of tests that take most of that 5-6 minutes and it would be good to identify those tests and redesign them to be faster (likely by changing the data that is used). 

As I recall the SDP solver tests are among the worst and there were some MINLP tests (i.e. juniper) that could use stream lining.

odow:
> that there are a small-ish number of tests that take most of that 5-6 minutes
> As I recall the SDP solver tests are among the worst

I've been doing some profiling locally and in #962, and this is still the case.

odow:
One contributor is

https://github.com/lanl-ansi/PowerModels.jl/blob/69be1eba2200f3c8f7c345fb7e1628989a60ce6e/test/multinetwork.jl#L225-L242

which takes up ~30 seconds of a 6 minute run.

But it's essentially all precompilation:
```julia
Test Summary:                                             | Pass  Total     Time
PowerModels                                               |  314    314  1m47.1s
  multinetwork.jl                                         |  314    314  1m47.1s
    test multinetwork                                     |  314    314  1m46.3s
      idempotent unit transformation                      |    2      2     9.2s
      topology processing                                 |   74     74     5.2s
      test solve_opf with multinetwork data               |    1      1     2.6s
      test solve_mn_opf with single-network data          |    1      1     0.1s
      test multi-network solution                         |    3      3    13.7s
      test make_multinetwork                              |    3      3     0.6s
      2 period 5-bus asymmetric case                      |   28     28    37.0s
        test dc polar opb                                 |    4      4     1.8s
        test ac polar opf                                 |    4      4     0.2s
        test dc polar opf                                 |    4      4     1.5s
        test soc opf                                      |    4      4     2.7s
        test sdp with constraint decomposition opf        |    4      4    27.6s
        test sdp with constraint decomposition opf REPEAT |    4      4     1.5s
        test nfa opf                                      |    4      4     1.7s
      2 period 5-bus dual variable case                   |   78     78     2.6s
      hybrid network case - polar                         |    6      6     4.1s
[...]

PR #962, opened 2025-04-22
Refactor test/runtests.jl and measure runtime

odow:
Part of #73

I want to see where we're spending time in the tests.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,80 +1,57 @@
 using PowerModels
+using Test
+
+import HiGHS
 import InfrastructureModels
+import Ipopt
+import JSON
+import JuMP
+import Juniper
+import LinearAlgebra
 import Memento
+import SCS
+import SparseArrays
 
 # Suppress warnings during testing.
 Memento.setlevel!(Memento.getlogger(InfrastructureModels), "error")
 PowerModels.logger_config!("error")
 
-import HiGHS
-import Ipopt
-import SCS
-import Juniper
-
-import JuMP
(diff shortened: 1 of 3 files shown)


ccoffrin:
I only happen to glance at this PR quickly, but I don't approve of some of these changes. We can discuss more at our meeting tomorrow.  In general, I think you should wait for my review and approval on all changes staged for master.

odow:
Ah okay :+1: shall I revert it?

test/runtests.jl
```diff
@@ -1,80 +1,57 @@
...
-
-    include("docs.jl")
-
+@testset verbose=true "PowerModels" begin
+    @testset verbose=true "$file" for file in readdir(@__DIR__)
+        if !endswith(file, ".jl") || file in ("common.jl", "runtests.jl")
+            continue
+        end
+        include(joinpath(@__DIR__, file))
+    end
```
ccoffrin, line 56: I prefer the explicit list of imports for the development workflow (which may be outdated), lets discuss that and if this this last change should be reverted.

odow: I've found that explicit list of importing test files is very brittle. It's quite easy to miss including test file, and it can encourage people to run test files independently, which can lead to dependent state between the test files.

I would be in favor of moving to something like the JuMP tests, which strongly encapsulate tests into modules and functions to remove all shared state: https://github.com/jump-dev/JuMP.jl/blob/master/test/test_expr.jl. It's quite a large change though.

PR #963, opened 2025-04-23
Remove empty docstrings

odow:
These are actually less helpful than nothing.

Before

```julia
help?> build_mn_opf
search: build_mn_opf build_mn_opf_strg build_mn_opf_bf_strg



```

Now

```Julia
help?> build_mn_opf
search: build_mn_opf build_mn_opf_strg build_mn_opf_bf_strg

  No documentation found.

  PowerModels.build_mn_opf is a Function.

  # 1 method for generic function "build_mn_opf" from PowerModels:
   [1] build_mn_opf(pm::AbstractPowerModel)
       @ ~/.julia/dev/PowerModels/src/prob/opf.jl:54
```


odow:
Ah okay. Then I should change the doc build

PR #966, opened 2025-04-24
Fix doc build and change warnings into errors

odow:
Partial revert of #963. Instead of empty docstrings, we now have a minimal signature. There are open issues to actually improve the documentation.

I've removed `warnonly = Documenter.except(:linkcheck),`, since it actually meant that these missing docstrings were warning instead of erroring.


ccoffrin:
these changes LGTM

PR #969, opened 2025-05-01
Remove epigraph formulation of quadratic objectives for ConicModel

### LINKED ISSUE #55: Remove Auxiliary Variables in Conic Objective
Update, objective_min_fuel_cost{T <: AbstractConicPowerFormulation}(pm::GenericPowerModel{T}) in objective.jl.

The key is to define the min on a norm form.

odow:
Closes https://github.com/lanl-ansi/PowerModels.jl/issues/55

We might need to test this on a much larger problem, since there are some test differences. I'll comment in-line on some things.

### DIFF
--- a/src/core/objective.jl
+++ b/src/core/objective.jl
@@ -224,60 +224,6 @@ function _pwl_cost_expression(pm::AbstractPowerModel, x_list, points; nw=0, id=1
 end
 
 
-
-# note that `cost_terms` should be providing in ascending order (the reverse of the Matpower spec.)
-function _polynomial_cost_expression(pm::AbstractPowerModel, x_list::Array{JuMP.VariableRef}, cost_terms; nw=0, id=1, var_name="x")
-    x = sum(x_list)
-    if length(cost_terms) == 0
-        return 0.0
-    elseif length(cost_terms) == 1
-        return cost_terms[1]
-    elseif length(cost_terms) == 2
-        return cost_terms[1] + cost_terms[2]*x
-    elseif length(cost_terms) == 3
-        return cost_terms[1] + cost_terms[2]*x + cost_terms[3]*x^2
-    else # length(cost_terms) >= 4
-        cost_nl = cost_terms[4:end]
-        return JuMP.@expression(pm.model, cost_terms[1] + cost_terms[2]*x + cost_terms[3]*x^2 + sum( v*x^(d+2) for (d,v) in enumerate(cost_nl)) )
-    end
-end
-
-# note that `cost_terms` should be providing in ascending order (the reverse of the Matpower spec.)
-function _polynomial_cost_expression(pm::AbstractConicModels, x_list::Array{JuMP.VariableRef}, cost_terms; nw=0, id=1, var_name="x")
-    x = sum(x_list)
-    if length(cost_terms) == 0
-        return 0.0
-    elseif length(cost_terms) == 1
-        return cost_terms[1]
-    elseif length(cost_terms) == 2
-        return cost_terms[1] + cost_terms[2]*x
-    elseif length(cost_terms) == 3
-        x_lb = sum(JuMP.lower_bound.(x_list))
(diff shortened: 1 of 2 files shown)


src/core/objective.jl
```diff
@@ -224,60 +224,6 @@ function _pwl_cost_expression(pm::AbstractPowerModel, x_list, points; nw=0, id=1
...
-        return cost_terms[1]
-    elseif length(cost_terms) == 2
-        return cost_terms[1] + cost_terms[2]*x
-    elseif length(cost_terms) == 3
-        return cost_terms[1] + cost_terms[2]*x + cost_terms[3]*x^2
-    else # length(cost_terms) >= 4
-        cost_nl = cost_terms[4:end]
-        return JuMP.@expression(pm.model, cost_terms[1] + cost_terms[2]*x + cost_terms[3]*x^2 + sum( v*x^(d+2) for (d,v) in enumerate(cost_nl)) )
-    end
-end
```
odow, line 243: This method, AFAIK, can just be removed in favor of the other that remains, regardless of what we choose here.

test/opf.jl
```diff
@@ -642,17 +642,19 @@ end
         result = solve_opf("../test/data/matpower/case6.m", SOCWRConicPowerModel, sdp_solver)
 
         @test result["termination_status"] == OPTIMAL
-        #@test isapprox(result["objective"], 11472.2; atol = 3e0)
-        #@test isapprox(result["objective"], 11451.5; atol = 3e0)
-        @test isapprox(result["objective"], 11473.4; atol = 3e0)
+        # Solvers get a range of answers for this test.
+        #  * Mosek@10 produces 11472.3
+        #  * SCS@3    produces 11470.3
+        @test 11470.2 <= result["objective"] <= 11472.4
```
odow, line 648: Previously, SCS reported 11473.4. In both cases, Mosek reports 11472.3. SCS's accuracy is ~1e-4, so both values are within tolerance.

test/opf.jl
```diff
@@ -642,17 +642,19 @@ end
...
         result = solve_opf("../test/data/matpower/case24.m", SOCWRConicPowerModel, sdp_solver)
 
         @test result["termination_status"] == OPTIMAL
-        #@test isapprox(result["objective"], 70693.9; atol = 1e0)
-        #@test isapprox(result["objective"], 70670.0; atol = 1e0)
-        @test isapprox(result["objective"], 70683.5; atol = 1e0)
+        # Solvers get a range of answers for this test.
+        #  * Mosek@10 produces 70690.7
+        #  * SCS@3    produces 70680.6
+        @test 70680.5 <= result["objective"] <= 70690.8
```
odow, line 657: Quite a broad range of possible "solutions" here

src/core/objective.jl
ccoffrin: Yes, this looks like a work around for an issue that is no longer needed.  It might be worth double checking some of the dependent packages (e.g., PowerModelsAnnex).

If something breaks may need to put back along with a test that checks this control path.

test/opf.jl
```diff
@@ -944,11 +946,14 @@ end
...
-        #@test isapprox(result["objective"], 6735.17; atol = 1e0)
-        @test isapprox(result["objective"], 6827.71; atol = 1e0)
+        # SCS hits an iteration limit with these, but it still finds the correct
+        # objective.
+        @test result["termination_status"] == OPTIMAL ||
+              result["termination_status"] == ITERATION_LIMIT
+        # Solvers get a range of answers for this test.
+        #  * Mosek@10 produces 6827.48
+        #  * SCS@3    produces 6827.73
+        @test 6827.38 <= result["objective"] <= 6827.83
```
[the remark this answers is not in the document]
ccoffrin: As long as this does not increase the test runtime a lot this is OK by me.

ccoffrin:
I am thinking it is best to put this in the next breaking release, in an abundance of caution.

Issue #697, opened 2020-04-13
AC Power Flow Improvements


ccoffrin:
Performance
- [ ] eliminate memory allocation needed by `inject` in warm-start feature
- [ ] explore use of complex numbers in PowerFlowData struct
- [ ] can `calc_admittance_matrix_inv` be improved by using a @view?
- [ ] can `injection_factors_va` be improved by using a @view?

Scope
- [ ] a support DC lines as a fixed injection?
- [ ] topology processing to support switches?

odow:
See 
https://github.com/lanl-ansi/PowerModels.jl/blob/685fbb260fe1a773aef2a89dc205e0525b42a030/src/core/admittance_matrix.jl#L187-L188

odow:
@ccoffrin there are zero tests that hit `calc_admittance_matrix_inv`?

And if I do
```Julia
using PowerModels
data = PowerModels.parse_file("../test/data/matpower/case24.m")
am = calc_admittance_matrix(data)
```
the admittance matrix is `ComplexF64`, so is `am_inv = zeros(Float64` intentional?

PR #970, opened 2025-05-01
Add tests for calc_admittance_matrix_inv

odow:
x-ref https://github.com/lanl-ansi/PowerModels.jl/issues/697

### DIFF
--- a/src/core/admittance_matrix.jl
+++ b/src/core/admittance_matrix.jl
@@ -174,7 +174,7 @@ function calc_susceptance_matrix_inv(data::Dict{String,<:Any})
 end
 
 "calculates the inverse of the susceptance matrix"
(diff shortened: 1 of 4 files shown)


ccoffrin:
Thanks!  LGTM.

src/core/admittance_matrix.jl
```diff
@@ -174,7 +174,7 @@ function calc_susceptance_matrix_inv(data::Dict{String,<:Any})
 end
 
 "calculates the inverse of the susceptance matrix"
```
odow, line 176:
```suggestion
"calculates the inverse of the admittance matrix"
```

odow:
Add docstrings to https://github.com/lanl-ansi/PowerModels.jl/blob/685fbb260fe1a773aef2a89dc205e0525b42a030/docs/src/power-flow.md?plain=1#L156-L159

PR #986, opened 2025-10-16
Update to JSON@1

odow:
Closes #985

@ccoffrin there's one notable change, with how JSON@1 writes `Inf`:

```julia
(js) pkg> st
Status `/private/tmp/js/Project.toml`
⌃ [682c06a0] JSON v0.21.4
Info Packages marked with ⌃ have new versions available and may be upgradable.

julia> import JSON

julia> JSON.json(Inf)
"null"

julia> JSON.json(-Inf)
"null"
```

```julia
(json) pkg> st
Status `/private/tmp/json/Project.toml`
  [682c06a0] JSON v1.1.0

julia> import JSON

julia> JSON.json(Inf)
ERROR: ArgumentError: Inf not allowed to be written in JSON spec; pass `allownan=true` to allow anyway
Stacktrace:
 [1] infcheck(x::Float64, allownan::Bool)
   @ JSON ~/.julia/packages/JSON/jLnej/src/write.jl:694
 [2] _number(buf::Vector{UInt8}, pos::Int64, x::Float64, opts::JSON.WriteOptions{JSON.JSONWriteStyle}, io::Nothing, bufsize::Int64)
   @ JSON ~/.julia/packages/JSON/jLnej/src/write.jl:712
 [3] json!
   @ ~/.julia/packages/JSON/jLnej/src/write.jl:592 [inlined]
 [4] json! (repeats 3 times)
   @ ~/.julia/packages/JSON/jLnej/src/write.jl:562 [inlined]
 [5] json(x::Float64; pretty::Bool, kw::@Kwargs{})
   @ JSON ~/.julia/packages/JSON/jLnej/src/write.jl:433
 [6] json(x::Float64)
   @ JSON ~/.julia/packages/JSON/jLnej/src/write.jl:426
 [7] top-level scope
   @ REPL[3]:1

julia> JSON.json(Inf; allownan=true)
"Infinity"

julia> JSON.json(-Inf; allownan=true)
"-Infinity"
```

It doesn't look like we have a test that parses a file containing `Infinity`. Does that ever come up in a MATPOWER file?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ PowerModels.jl Change Log
 =========================
 
 ### Staged
+- Update to JSON@1 (#986)
 
 ### v0.21.5
 - Relax tests to allow `INFEASIBLE_POINT` (#976)
--- a/Project.toml
+++ b/Project.toml
@@ -1,8 +1,8 @@
 name = "PowerModels"
 uuid = "c36e90e8-916a-50a6-bd94-075b64ef4655"
+version = "0.21.5"
 authors = ["Carleton Coffrin"]
 repo = "https://github.com/lanl-ansi/PowerModels.jl"
-version = "0.21.5"
 
 [deps]
 InfrastructureModels = "2030c09a-7f63-5d83-885d-db604e0e9cc0"
@@ -18,7 +18,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 HiGHS = "1"
 InfrastructureModels = "0.6, 0.7"
 Ipopt = "1"
-JSON = "0.21"
+JSON = "1"
 JuMP = "1.15"
 Juniper = "0.9"
 Memento = "1"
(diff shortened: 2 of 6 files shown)


odow:
One argument against parsing with `allownan = true` is this bug: https://github.com/JuliaIO/JSON.jl/issues/397

```Julia
julia> import JSON

julia> JSON.parse("1"; allownan = true)
1.0

(json) pkg> st
Status `/private/tmp/json/Project.toml`
  [682c06a0] JSON v1.1.0
```

ccoffrin:
As i recall the convention in Matpower is that “0” some times means “Inf”
or “skip”.

I need to carefully think about the best way to proceed.  Maybe we can have
a call to discuss in 2 weeks?

odow:
Maybe the safest thing to do is just to revert to the old behaviour, which we can do with:

```julia
julia> JSON.json(Inf; allownan = true, inf="null")
"null"
```

Then there's no breaking change.

The best long-term behaviour is still up for debate, but no one seems to have complained about it so let's stick with it for now?

ccoffrin:
that works for me.

PR #997, opened 2026-04-04
Use Logging instead of Memento and bump Julia requirement to 1.10

### LINKED ISSUE #996: Memento dependency is not fully Julia 1.13 compatible
It may be worth reconsidering having Memento as dependency. It is depreciated package (last updated in 2022). On Julia 1.13 beta-3 I get following during precompile:

```
┌ Memento
│  WARNING: Method definition var"@test_throws"(LineNumberNode, Module, Any, Any, Any) in module Test at C:\workdir\usr\share\julia\stdlib\v1.13\Test\src\Test.jl:876 overwritten in module TestUtils at C:\Users\karlo\.julia\packages\Memento\xnHxE\src\memento_test.jl:119.
│  ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
└
```

KSepetanc:
Fixes #996 and bumps Julia requirement to 1.10 (1.8 is required for logging, but 1.10 is in line with recent decision for InfrastructureModels.jl to go to LTS).


odow:
I probably need to discuss this with the LANL folks, because a lot of their code uses Memento. It would be easier if we just updated Memento.

KSepetanc:
I can swap it in InfrastructureModels.jl too if decided so.

ccoffrin:
Switching from Memento to Julia's builtin Logging has been a desire since the release of Julia v1.0.  Early attempts failed because we could not figure out how to make separate logging streams for each infrastructure models package.  It is a hard requirement that we can turn these on and off selectively.  I think PowerModelsDistribution did make the jump, but I was not involved in the design and it did not proliferate to the other InfrastructureModels packages.

Even if moving from Memento to Logging is be desirable, it could be a major code refactor that would touch a lot of packages.  We should way that carefully against all alternatives, including make a minor patch to Memento for version compatibility, as @odow suggests.

KSepetanc:
Sounds good. Probably we will need to change Julia registry to point to patched Memento as we do not have merge rights and there are no maintainers anymore (I leave that to your side). Furthermore, I can take a look at how PowerModelsDistribution resolved separate logging issue, and if applicable, I propose we swap Memento for Logging too when ready.

KSepetanc:
Implemented per module logging just like it is in PowerModelsDistribution. LoggingExtras is used for it. Added also unit tests. Marking as ready for review.

You may also want to fix one unrelated (solver tolerance) test failure on Julia 1.13 in another PR:
`failure is in test/opf.jl:439 — a numerical tolerance issue in the "24-bus rts case" test (82242.64 vs 82240, off by ~2.6). This is a pre-existing solver tolerance issue unrelated to our logging changes.`

odow:
I'll talk to Frames

skylerreid:
I ported this change over to GasModels and have a working version there now. It looks like this should be a pretty smooth change across the suite of packages (I just used find and replace plus @KSepetanc 's `src/core/logging.jl` file). That PR is linked [here](https://github.com/lanl-ansi/GasModels.jl/pull/303)

odow:
Okay, great. Let me review this in more detail. I also talked to Carleton, and even if we could fix Memento this once, it's probably easier in the long term if we remove it.

src/PowerModels.jl
```diff
@@ -5,28 +5,31 @@ import InfrastructureModels: optimize_model!, @im_fields, nw_id_default
...
+_DEFAULT_LOGGER = Logging.current_logger()
+_LOGGER = Logging.ConsoleLogger(; meta_formatter=PowerModels._pm_metafmt)
+function __init__()
+    global _DEFAULT_LOGGER = Logging.current_logger()
+    global _LOGGER = Logging.ConsoleLogger(; meta_formatter=PowerModels._pm_metafmt)
 
-# Register the module level logger at runtime so that folks can access the logger via `getlogger(PowerModels)`
-# NOTE: If this line is not included then the precompiled `PowerModels._LOGGER` won't be registered at runtime.
-__init__() = Memento.register(_LOGGER)
+    Logging.global_logger(_LOGGER)
```
odow: I need to think about this. I don't think it's viable. This mutates global state for all Julia packages and will conflict if there are multiple InfrastructureModels trying to do the same thing. We need a per-package (or per-module) logger.

odow:
I've been working on an alternative in https://github.com/lanl-ansi/PowerModels.jl/pull/1000

odow:
Okay, so I think I understand this PR better (please correct me if I'm wrong):

Let's assume we are using both this PR to PowerModels and https://github.com/lanl-ansi/InfrastructureModels.jl/pull/98. The flow is:

* InfrastructureModels runs its `__init__` and sets the global logger to `Logging.ConsoleLogger(; meta_formatter)`
* PowerModels runs its `__init__` and overrides the global logger to its version of `Logging.ConsoleLogger(; meta_formatter)`
* Someone does `InfrastructureModels.silence!()`
* InfrastructureModels creates an `LoggingExtras.EarlyFilteredLogger` that wraps PowerModel's `Logging.ConsoleLogger`.
* Someone does `PowerModels.logger_config!(:warn)`. This wraps a new `EarlyFilteredLogger` around InfrastructureModel's `EarlyFilteredLogger`.

Here's an example where it can go wrong with the current PR (yes, we can fix it, but it goes to show that the logic is non-trivial):
```julia
julia> import PowerModels

julia> PowerModels._IM.build_solution_values(:a)
┌ InfrastructureModels | Warning ] : build_solution_values found unknown type Symbol
└ @ InfrastructureModels ~/git/lanl-ansi/InfrastructureModels/src/core/solution.jl:169
:a

julia> PowerModels.parse_pti("test/data/pti/parser_test_c.raw")
┌ PowerModels | Warning ] : Version 32 of PTI format is unsupported, parser may not function correctly.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:695
┌ PowerModels | Warning ] : At line 4, new section started with '0', but additional non-comment data is present. Pattern '^\s*0\s*[/]*.*' is reserved for section start/end.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:655
Dict{String, Array{Dict}} with 1 entry:
  "CASE IDENTIFICATION" => [Dict{String, Any}("IC"=>0, "BASFRQ"=>60.0, "NXFRAT"=>1, "Comment_Line_1"=>"…

julia> PowerModels._IM.silence!()

julia> PowerModels.parse_pti("test/data/pti/parser_test_c.raw")
┌ PowerModels | Warning ] : Version 32 of PTI format is unsupported, parser may not function correctly.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:695
┌ PowerModels | Warning ] : At line 4, new section started with '0', but additional non-comment data is present. Pattern '^\s*0\s*[/]*.*' is reserved for section start/end.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:655
Dict{String, Array{Dict}} with 1 entry:
  "CASE IDENTIFICATION" => [Dict{String, Any}("IC"=>0, "BASFRQ"=>60.0, "NXFRAT"=>1, "Comment_Line_1"=>"…

julia> PowerModels._IM.build_solution_values(:a)
:a

julia> PowerModels.silence!()

julia> PowerModels._IM.reset_logging_level!()

julia> PowerModels.parse_pti("test/data/pti/parser_test_c.raw")
┌ PowerModels | Warning ] : Version 32 of PTI format is unsupported, parser may not function correctly.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:695
┌ PowerModels | Warning ] : At line 4, new section started with '0', but additional non-comment data is present. Pattern '^\s*0\s*[/]*.*' is reserved for section start/end.
└ @ PowerModels ~/git/lanl-ansi/PowerModels/src/io/pti.jl:655
Dict{String, Array{Dict}} with 1 entry:
  "CASE IDENTIFICATION" => [Dict{String, Any}("IC"=>0, "BASFRQ"=>60.0, "NXFRAT"=>1, "Comment_Line_1"=>"…
```
Uh oh, even though we said `PowerModels.silence!`, `reset_logging_level!` reset the logger to the one that happened _before_ `IM.silence!()`.

There's also the problem that loading InfrastructureModels overrides _all_ logging, even in unrelated packages:
```julia
julia> @warn("Hello")
┌ Main | Warning ] : Hello
└ @ Main REPL[11]:1
```

This actually offers an alternative approach: we could write the logger once in InfrastructureModels, and then all downstream packages would have the same logging.

We could fix the EarlyFilteredLogger issue by a package-level global constant that we check instead of adding and removing the logging layers.

The approach in #1000 is to explicitly have a package-dependent logger, with the downside that it isn't composable with other loggers, like the one in `Test.@test_logs`.

Tricky, tricky. It's not super clear what the best approach is.

src/core/admittance_matrix.jl
```diff
@@ -23,10 +23,10 @@ Base.show(io::IO, x::AdmittanceMatrix{<:Number}) = print(io, "AdmittanceMatrix($
 "data should be a PowerModels network data model; only supports networks with exactly one reference bus"
 function calc_admittance_matrix(data::Dict{String,<:Any})
     if length(data["dcline"]) > 0
-        Memento.error(_LOGGER, "calc_admittance_matrix does not support data with dclines")
+        error("calc_admittance_matrix does not support data with dclines")
```
odow: This is also not exactly the same. It errors but it doesn't write the same `grep`able log message.

KSepetanc:
Previously, PowerModelsDistribution approach was replicated.

Pardon me for heavy ai usage. I do this in free time after work. Just ignore if not worthy.

Tests fail due to IM change not being merged.

Here is ai summary:

> Single global logger owner — Only InfrastructureModels calls global_logger() in __init__(). Downstream packages register via _IM.register_module!() and never touch the global logger, fixing the "last __init__ wins" conflict.
> 
> Independent per-module filtering — Log level overrides are stored in a shared Dict{Module, LogLevel} checked by a single EarlyFilteredLogger. PowerModels.silence!() and InfrastructureModels.silence!() are fully independent — no more nested logger stacking or state corruption on reset.
> 
> @log_error macro — Restores Memento.error() behavior: emits a grepable log record with correct caller file/line before throwing ErrorException. All 109 former Memento.error call sites now use this, addressing the "not the same grepable log message" comment on admittance_matrix.jl.
> 
> LoggingExtras removed from PowerModels deps — Only IM needs it now. PM's logging.jl is a thin wrapper delegating to IM's shared API.

odow:
@KSepetanc did you use AI to make the original PR? It's just a straight copy from PowerModelsDistribution:

 * https://github.com/lanl-ansi/PowerModelsDistribution.jl/pull/336
 * https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/main/src/core/logging.jl

Edit: Nevermind, I guess I missed your comment in https://github.com/lanl-ansi/PowerModels.jl/pull/997#issuecomment-4189350531

PR #928, opened 2024-10-08
Support matpower dict

yasirroni:
This code support `PowerModels` to accept `matlab_data` in the form of `Dict`.

But, to be honest, later in the next PR, I want it to accept `Dict{Any, Any}` and `InfractructureModels` to accept `AbstractMatrix`.

Here is current usage:

```julia
    function convert_matrix_to_vector_of_reals(value::Any)
        if isa(value, AbstractMatrix)
            return [Vector(row) for row in eachrow(value)]
        else
            return value 
        end
    end

    function convert_dict_value_matrix_to_vector_of_reals(data::Dict)
        for (key, value) in data
            data[key] = convert_matrix_to_vector_of_reals(value)
        end
        return data
    end

    function convert_python_to_matlab_data(dict::Dict{Any, Any})
        new_dict = Dict{String, Any}()
        for (key, value) in dict
            new_dict[string(key)] = convert_matrix_to_vector_of_reals(value)
        end
        return new_dict
    end

    matlab_data = convert_python_to_matlab_data(matlab_data)
    data = PowerModels.parse_matpower(matlab_data, func_name=func_name, colnames=colnames, validate=true)
    result = solve_opf(data, ACPPowerModel, nlp_solver)
```

Where `matlab_data` id Dict{Any, Any}` that contains `Matrix{Float64}`.

PR #998, opened 2026-04-05
Expose `_parse_matlab_data` by moving `_IM.parse_matlab_string` outside

yasirroni:
### Change

Splits `_parse_matpower_string` into two internal functions with **zero logic changes**:
```julia
# Before
function _parse_matpower_string(data_string::String)
    matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)
    # ... all processing logic
end

# After
function _parse_matpower_string(data_string::String)
    matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)
    return _parse_matlab_data(matlab_data, func_name, colnames)
end

function _parse_matlab_data(matlab_data::Dict{String,Any}, func_name::Union{AbstractString,Nothing}, colnames::Dict{String,Any})
    # ... exact same body, untouched
end
```

### What is NOT changed

- `_parse_matpower_string` behavior is identical
- `parse_matpower(io::IO)` and `parse_matpower(file::String)` are identical
- No new public API in this PR
- No new data format, no new dependencies

### Why it matters

This allows downstream packages to wrap PowerModels without serializing back to a `.m` string. For example, someone with matpower dict data use a wrapper that **live outside** PowerModels:

```julia
function parse_matpower(matlab_data::Dict{String,Any};
                        func_name::Union{AbstractString,Nothing}=nothing,
                        colnames::Dict{String,Any}=Dict{String,Any}(),
                        validate=true)::Dict
    mp_data = PowerModels._parse_matpower_data(matlab_data, func_name, colnames)
    pm_data = PowerModels._matpower_to_powermodels!(mp_data)
    if validate
        PowerModels.correct_network_data!(pm_data)
    end
    return pm_data
end
```

As suggested in https://github.com/lanl-ansi/PowerModels.jl/pull/928, this PR will support a wrapper lives outside PowerModels, without rewriting the whole `_parse_matpower_string`, thus safer if in the future someone update `_parse_matpower_string`. This PR will make minimal changes and minimal burden to support this.

___

Alternatively, at least for me, actually, it would be cleaner if the refactor is:

```julia
function parse_matpower(io::IO; validate=true)::Dict
    data_string = read(io, String)
    matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)
    mp_data = _parse_matlab_data(matlab_data, func_name, colnames)
    pm_data = _matpower_to_powermodels!(mp_data)
    if validate
        correct_network_data!(pm_data)
    end
    return pm_data
end
```

But, this alternative refactor remove `_parse_matpower_string`.

### DIFF
--- a/src/io/matpower.jl
+++ b/src/io/matpower.jl
@@ -125,7 +125,11 @@ const _mp_switch_columns = [
 
 function _parse_matpower_string(data_string::String)
     matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)
+    return _parse_matpower_data(matlab_data, func_name, colnames)
+end
 
+
+function _parse_matpower_data(matlab_data::Dict{String,Any}, func_name::Union{AbstractString,Nothing}, colnames::Dict{String,Any})
     case = Dict{String,Any}()
 
     if func_name != nothing
--- a/src/io/matpower.jl
+++ b/src/io/matpower.jl
@@ -125,11 +125,11 @@ const _mp_switch_columns = [
 
 function _parse_matpower_string(data_string::String)
     matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)
-    return _parse_matpower_data(matlab_data, func_name, colnames)
+    return _parse_matlab_data(matlab_data, func_name, colnames)
 end
 
 
-function _parse_matpower_data(matlab_data::Dict{String,Any}, func_name::Union{AbstractString,Nothing}, colnames::Dict{String,Any})
+function _parse_matlab_data(matlab_data::Dict{String,Any}, func_name::Union{AbstractString,Nothing}, colnames::Dict{String,Any})
     case = Dict{String,Any}()
 
     if func_name != nothing


odow:
> This allows downstream packages 

Which packages?

> to wrap PowerModels without serializing back to a .m string

Why is this important?

(A micro benchmark showing that this is slow is not a justification. You would need a benchmark showing an actual end use case here reading and writing files is a bottleneck. The true bottleneck is almost certainly the solver.)

Also note that functions starting with `_` are not public. We reserve the right to change or remove them in any future release. You should not use them in down-stream packages.

yasirroni:
> Which packages?

I've not use or create one yet, but the idea is to use dict instead of string of matpower data.

> Why is this important?

So that I don't need to write .m file. I can just pass dict to PowerModels. And dict format is easier to achieve if the data passed is from Python, for example. This is not about speed, just make running PowerModels from Python for example, easier (without reading/writing string, but passing data directly).

> Also note that functions starting with _ are not public. We reserve the right to change or remove them in any future release. You should not use them in down-stream packages.

I understand.

yasirroni:
The most reasonable use case it to run case69.m that is currently not support by PowerModels.

https://github.com/lanl-ansi/PowerModels.jl/pull/928#issuecomment-2439745625

With this PR, one can run, for example, python code, run matlab engine, call matpower, read matpower data, pass to python, pass to julia, pass to powermodels. Yeah, this is cumbersome, but at least supported, compared to impossible (or rewrite the .m file into powermodels data manually).

odow:
This feels like an https://en.wikipedia.org/wiki/XY_problem.

Why Python, MATLAB, and Julia?

Why is writing out a new `.m` file hard?

PowerModels also has a JSON interface: https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/io/json.jl. That might be a better file format from Python?

yasirroni:
> Why Python, MATLAB, and Julia?

PowerModels run on Julia. The main idea is to use native .m file reader, that is matlab. I don't know if there is matlab/julia bridge tools. Thus I go around matlab/python/julia.

> Why is writing out a new .m file hard?

Writing by hand, is out of option.
Writing "string" as the way to store a modified file could potentially introduce bug.

>  PowerModels also has a JSON interface: https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/io/json.jl. That might be a better file format from Python?

Yes, this is indeed better, for data already in PowerModels json data format. For data that is still in matpower json format, if you want to support `_parse_matpower_string` or `_parse_matlab_data`, we go back to our original problem.

___

This PR seriously didn't want to mess with anything, just move `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)` to elsewhere so that I can use all the other code in `_parse_matpower_string(data_string::String)` that is after `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)`.

odow:
> I don't know if there is matlab/julia bridge tools

Have you seen https://github.com/JuliaInterop/MATLAB.jl?

> to elsewhere so that I can use all the other code in _parse_matpower_string(data_string::String) that is after matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true).

I'm very hesitant to merge this PR (even though I agree it is a small change) because currently because these functions are private. We'd need to agree to test them, document them, and maintain them going forward.

My main concern is that there might be a better way to achieve what you are trying to do without needing this PR.

yasirroni:
> Have you seen https://github.com/JuliaInterop/MATLAB.jl?

This is my first time. But still, after we read the .m files with matpower/matlab/julia, we (might) be ended up with dict, struct, or something similar. If I want to convert this into power models format, it would be easier to just pass it into `_parse_matlab_data` instead of, once again, I rewrite the `_parse_matlab_string` on my code plus removing the `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)`.

But, if you guys still insist to not want to move `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)`, I will just rewrite, copy paste, everything you have under `_parse_matlab_string`.

> I'm very hesitant to merge this PR (even though I agree it is a small change) because currently because these functions are private. We'd need to agree to test them, document them, and maintain them going forward.

Once again, this PR didn't change any API already exposed for user. It is just me, using internal function that should not be used, but I do like to take the risk, so that you don't need to test or change anything in the test.

Except, if you guys want to make everything public, that is  `_parse_matpower_string`, `_parse_matpower_data`, and even `_matpower_to_powermodels`, I would be very happy.

skylerreid:
GasModels maintainer here: I'd also be a little hesitant to start messing with some of the private parser APIs. If you're looking for something that's highly compatible across Julia/Matlab/Python, just use JSON instead of `.m`. 

I've done a lot of work on the current GMs parser and this looks like a solution in search of a problem.

yasirroni:
> just use JSON instead of .m

That is the spirit. Reading json, Python and Julia both use dict. Converting MATLAB struct to Python, we also got dict. Passing Python dict to Julia, we got dict. That is why, just move `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)` away so that I can use PowerModels handling dict (the `matlab_data` out put of `matlab_data, func_name, colnames = _IM.parse_matlab_string(data_string, extended=true)` is dict).

Dict is the way. Limiting to string input is not convenient.

___

Else, you want to enforce everyone doing read and write a `.json` file. But still, PowerModels only support `.json` that is already structured as PowerModels case format.

skylerreid:
> you want to enforce everyone doing read and write a `.json` file

It's trivially easy to serialize a julia/python dict to json without writing it to a file, by using an IO stream. Julia and Python both have first-class support for this, and I think Matlab does too. If you want to be able to pass a dictionary to the parser from another language, that's almost certainly the easiest solution.

yasirroni:
> It's trivially easy to serialize a julia/python dict to json without writing it to a file, by using an IO stream. Julia and Python both have first-class support for this, and I think Matlab does too.

But the data type is a dictionary in Python, right? I used PyCall some time ago, and it used dictionaries on both sides. Julia also natively parses JSON into dictionaries.

> If you want to be able to pass a dictionary to the parser from another language, that's almost certainly the easiest solution.

That is what I wanted to *partially* support with this PR: simply allow translating a dictionary in "MATPOWER" structure to a dictionary in "PowerModels" structure using the private API. Internally, this is what PowerModels already does anyway.

___

I understand that using a private API is risky, but making it a proper public API would increase maintenance burden.

Thus, it is up to all the maintaners, but we seem to agree that Dict/JSON is the right approach. However, transforming one JSON format/schema to another is still required. Fortunately, PowerModels already supports this through dictionary transformations within this PR.

This PR is not to support .m file, but to "improve" abstraction, separating string processor from dict processor.

skylerreid:
> But the data type is a dictionary in Python, right?

no, it's a string. you can turn a python dict into a string, and pass that directly to julia using the json library on each end. json strings are easy to handle in basically any modern language.

yasirroni:
> > But the data type is a dictionary in Python, right?
> 
> no, it's a string. you can turn a python dict into a string, and pass that directly to julia using the json library on each end. json strings are easy to handle in basically any modern language.

Ah that is what I mean, you do:

```python
with open("output.json", "w") as file:
    json.dump(data, file)
```

And then Julia will read that file as `Dict` object too. Right?

And both (before Python `dump` and Julia `parsefile`) will use "MATPOWER" structure. Thus, we need a converter to translate that (after read using Julia `parsefile`) to "PowerModels" structure. That is what this PR does, make user able to use:

```julia
function _parse_matlab_data(matlab_data::Dict{String,Any}, func_name::Union{AbstractString,Nothing}, colnames::Dict{String,Any})
    # ... exact same body, untouched
end
```

PR #1000, opened 2026-04-09
Remove Memento and switch to Logging.jl

### LINKED ISSUE #996: Memento dependency is not fully Julia 1.13 compatible

odow:
Closes #996
Closes #997

This took me a while to get right. I still need to think about it.

The main issue is that `@test_warn` doesn't work anymore, so I needed to hack around it. `@test_logs` doesn't work, because the test sets the global logger, but we ain't using that, so we need to intercept the actual `PowerModels._LOGGER`.


skylerreid:
This looks great, nice work. Should I get started on porting this over to GasModels?

KSepetanc:
Should we add check in macro to avoid allocations when silence! is on due to performance?

example:

```
macro _warn(msg)
    return :(Logging.Warn >= Logging.min_enabled_level(_LOGGER) && Logging.with_logger(() -> @warn($msg), _LOGGER)) |> esc
end
```

I will try to recode #997 approach to have another option (not necessary better than this).

odow:
I tried the InfrastructureModels one as well: https://github.com/odow/InfrastructureModels.jl/pull/1.

I'm not sure if we want the file and line number information. If we don't have that, then we could change the macros back to functions. They need to be macros only so that we preserve line number information.

KSepetanc:
Deleted in line comments as resolved.

odow:
Great. I think we're all agreed that this is the right approach. We can always iterate on the logging string after merge.

For now I've removed your line number stuff. We didn't have it before and it adds a lot of clutter. I'm open to adding it back in the future though.

KSepetanc:
Thanks @odow. Superb work. I am very happy that it works with only native `Logger.jl`. It is least dependencies. Also should run fast in silence mode with that `if` so Win-Win.

odow:
Merging. I'll sort the test failure in a different PR. It looks unrelated.

PR #1001, opened 2026-04-13
Relax flakey tests that are failing on macOS

odow:


KSepetanc:
Check out if it is the same as I observed on Windows and Julia 1.13 (previously reported):

> You may also want to fix one unrelated (solver tolerance) test failure on Julia 1.13 in another PR:
> failure is in test/opf.jl:439 — a numerical tolerance issue in the "24-bus rts case" test (82242.64 vs 82240, off by ~2.6). This is a pre-existing solver tolerance issue unrelated to our logging changes.

odow:
We're constantly firefighting tolerance issues. Many of the models are non-convex, and Ipopt tends to find slightly different solutions on different platforms and Julia versions.

PR #1011, opened 2026-07-27
Change compute_ac_pf iteration to use more standard (smaller) Jacobian

### LINKED ISSUE #1009: Newton-Raphson power flow convergence differs from other power flow tools
## Convergence difference with example:

I have been using `compute_ac_pf( )` extensively and noticed a difference between its' convergence and that of a comparable tool, `pandapower`, both of which (can) use the same algorithm (Newton-Raphson Power Flow) with the same configuration options. There are cases where `pandapower` converges while `PowerModels` does not.

Here is an example of a `PowerModels` network that works (converges) in `pandapower` but not in `PowerModels` (diverges with Newton Raphson/ oscillates with trust region):
[works_pp_doesnt_work_pm.json](https://github.com/user-attachments/files/30053861/works_pp_doesnt_work_pm.json)

And here is a very similar one that works with both (converges to the same solution):
[works_both.json](https://github.com/user-attachments/files/30053873/works_both.json)

(For an easy run of the two networks with the two tools, see [this repo](https://github.com/LKuhrmann/PowerModelsVSpandapower.git))

Both are derived from `case30.m`, with the generation and load increased, shunts added and some generators replaced with their "inverse load" and at their reactive limits.

Both use the same warm start.

## My suspicion for the difference in convergence:

`PowerModels`' internal Newton Raphson power flow Jacobian structure seems non-standard to me. 
The Jacobian is of `2n * 2n` size where `n` is the number of busses, including PQ and PV busses and the reference bus.
More standard power flow implementations have a size of `2j+k * 2j+k` where `j` = # PQ busses and `k` = # PV busses  and there is one additional reference bus.

For a three bus system, where bus 1 is the reference bus, bus 2 is a PQ bus and bus 3 is a PV bus, and there are branches between buses 1 and 2 and buses 2 and 3, the PowerModels.jl jacobian looks like this:

#### PowerModels jacobian 
```math
F = J \cdot \Delta x
```

```math
\begin{bmatrix} 
 \Delta P_1 \\
 \Delta Q_1\\
 \Delta P_2 \\
 \Delta Q_2\\
 \Delta P_3 \\
 \Delta Q_3
\end{bmatrix}
=

\begin{bmatrix} 
%d %d
1 & 0 &  0 & \frac{\partial P_3}{\partial \delta_3}  &  & \\
0 & 1 &  0 & \frac{\partial Q_3}{\partial \delta_3} &  & \\
 &  & 0  & \frac{\partial P_2}{\partial \delta_2} &   \frac{\partial P_2}{\partial V_3} & \frac{\partial P_2}{\partial \delta_3} \\
  & & 1  & \frac{\partial Q_2}{\partial \delta_2} &  \frac{\partial Q_2}{\partial V_3} & \frac{\partial Q_2}{\partial \delta_3} \\
  & &  0 &   \frac{\partial P_3}{\partial \delta_2} & \frac{\partial P_3}{\partial V_3} & \frac{\partial P_3}{\partial \delta_3}\\
   & &  0 &  \frac{\partial Q_3}{\partial \delta_2} &  \frac{\partial Q_3}{\partial V_3} & \frac{\partial Q_3}{\partial \delta_3}
\end{bmatrix}

\begin{bmatrix} 
 \Delta P_1\\
 \Delta Q_1\\
 \Delta P_2 \\
 \Delta \delta_2\\
 \Delta V_3 \\
 \Delta \delta_3
\end{bmatrix}

``` 
(This is my understanding of the internal workings of `_compute_ac_pf` in `pf.jl`)

#### More standard jacobian
In a more standard power flow implementation the matrix for the same three bus system looks like this:

```math

\begin{bmatrix} 
 \Delta Q_2\\
 \Delta P_3 \\
 \Delta Q_3
\end{bmatrix}

=


\begin{bmatrix} 
   \frac{\partial Q_2}{\partial \delta_2} &  \frac{\partial Q_2}{\partial V_3} & \frac{\partial Q_2}{\partial \delta_3} \\
    \frac{\partial P_3}{\partial \delta_2} & \frac{\partial P_3}{\partial V_3} & \frac{\partial P_3}{\partial \delta_3}\\
    \frac{\partial Q_3}{\partial \delta_2} &  \frac{\partial Q_3}{\partial V_3} & \frac{\partial Q_3}{\partial \delta_3}
\end{bmatrix}

\begin{bmatrix} 
 \Delta \delta_2\\
 \Delta V_3 \\
 \Delta \delta_3
\end{bmatrix}

``` 

There seem to be unnecessary additional elements (the active power injection at PV busses and the active and reactive injection at the reference bus) in the core iteration that dont need to be in the iteration (as you get them "for free" when evaluating the nodal power balance) which can (in some cases) lead to divergence.

When reducing this internal structure to the more standard implementation (see [this branch](https://github.com/LKuhrmann/PowerModels.jl/tree/smaller_Jacobian)), the convergence difference in the example disappears.

### My questions:
Why is the jacobian in PowerModels.jl structured the way it is? 
Does it have advantages?

From my perspective it seems like the power flow should be changed to be more in line with the standard approach.

Thanks for the great package!

LKuhrmann:
## Change
Fixes #1009.

Reduces what is iterated in the Newton Raphson iteration from [P, Q] for all buses to [P, Q] for PQ buses and [P] for PV buses. This reduces the Jacobian size by the number of PV busses +2. 

I believe the implementation I am suggesting is more standard. The first two random results I get when I google for "newton raphson power flow" support this: [random university pdf 1](https://web.engr.oregonstate.edu/~webbky/ESE470_files/Section%205%20Power%20Flow.pdf), page 111ff and [random university pdf 2](https://ocw.tudelft.nl/wp-content/uploads/PowerFlow.pdf), page 9

## Impact
Should make `compute_ac_pf` slightly faster and more memory efficient with more standard convergence when compared to other power flow tools.

As this PR simplifies the Newton-Raphson iteration by no longer converging toward Q injections (for PV and slack buses) and P injection (for the slack bus) I would expect this to converge whenever the current implementation converges. 
But I can't prove that, so there is a risk that this might break convergence in some (probably a bit weird) cases.

## Todo
I assume this change will need some tests. One test that could be added is the one presented in #1009 that does not currently converge.

I am uncertain if the way I implemented this is inline with the broader PowerModels codebase, so I am keen for code changes/opinions!

What else would need to happen so this can be merged?
If there are reasons for the current power flow implementation I am keen to hear them!


LKuhrmann:
Seems like the remaining failing tests would be fixed by #1010

langestefan:
I think this will be cleaner to implement on top of #1013. That PR also converges on your test case. 

This PR adds duplicate code to `admittance_matrix.jl`. It also adds new fields to `AdmittanceMatrix`, which is also used by packages such as PowerModelsACDC: https://github.com/Electa-Git/PowerModelsACDC.jl/blob/8b8752720d454bba26237eb3a8012f6467bf48bd/src/prob/sacdcpf.jl#L760

We can instead put the map in `PowerFlowData`, which I think will be cleaner.

LKuhrmann:
Interesting that #1013 converges on my test case, as it also seems to use the larger (nonstandard) Jacobian.

I don't really have an opinion on what would be cleaner or how newton-rahpson should be implemented, my primary concern is convergence parity with other newton-raphson solvers.

langestefan:
I am still investigating, but it seems we found a bug that you resolved in this PR here:

https://github.com/lanl-ansi/PowerModels.jl/pull/1011/changes#diff-2b965b9925b2c961928bd613df19e77854d7f5f85c49b24fbea67f20914245bbR587-R590

On the main branch it's taking the state vectors from the last call to `f!`. But that means that `OnceDifferentiable` gets the Jacobian before it computes the residual, meaning at every iterate the Jacobian we use is the one from the previous step. This is actually the thing that causes the divergences on some test cases you observed, not the Jacobian formulation. 

This is why my Newton implementation without the reduced Jacobian from #1012 also converges. Because it also does not make this mistake. We should make a new issue for this.

LKuhrmann:
Very interesting!

I will see if I can find other cases where the convergence difference is the result of the smaller jacobian rather than a result of the jacobian using the newest voltages.

If updating the voltage twice (as I implemented in the lines you link) fixes the convergence issues, then it could be a very simple bugfix from a convergence perspective.

langestefan:
I have opened #1016 to discuss.

